PR #31, opened 2023-08-25
Add OVMS metrics to UWM list

### LINKED ISSUE #13: Identify model metrics that will be part of the PoC.
### Use Story

- #6 

### A/C
- Needed metrics for the PoC are documented.

grdryn:
Closes #13

### DIFF
--- a/acm/odh-edge/acm-observability/files/uwm_metrics.yaml
+++ b/acm/odh-edge/acm-observability/files/uwm_metrics.yaml
@@ -1,5 +1,6 @@
   uwl_metrics_list.yaml: |
     names:
+      # MLServer metrics
       - parallel_request_queue_sum
       - parallel_request_queue_bucket
       - parallel_request_queue_count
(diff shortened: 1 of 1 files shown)


acm/odh-edge/acm-observability/files/uwm_metrics.yaml
```diff
@@ -8,3 +9,19 @@
       - rest_server_request_duration_seconds_count
       - rest_server_requests_in_progress
       - rest_server_requests_total
+
+      # OpenVINO metrics
```
devguyio, line 13: Do you think it'd be useful to add comments with the metric type shown [in this gist](https://gist.github.com/grdryn/3fccf4805273b9ec91c135215e3dd0b9)

grdryn: I don't _think_ that would be useful, since the source of truth for that information is the metrics endpoint (and gets pulled into the Prometheus server too, I think), and it's discoverable from there. If we add it statically here, then if there are changes, this would be out-of-date/incorrect.

I'm happy to add it if that's the consensus though.

StevenTobin (approved):

acm/odh-edge/acm-observability/files/uwm_metrics.yaml
devguyio: it's not a showstopper for me, and I understand your concern about the source of truth but you can also say the same thing about the metrics names.

I feel it'd help when creating the dashboards or just trying to understand the metrics without having a running service. Documenting metrics names and types is a pretty common practice. But as I said, it's not a show stopper and since this is merged you can put your efforts else where now.

grdryn: If it's a common practice, then we should copy the same pattern as the other places where it's common. If you can link one or more places as an example of which bits of information is expected and in what format, we can look to work towards that! :+1:

PR #37, opened 2023-08-29
Add ACM files for registration of odh-core clusters

LaVLaS:
## Description
Add the ACM registration files for `odh-core` cluster to standardize the onboarding of infrastructure using ACM

## How Has This Been Tested?
These files were manually deployed on the ACM hub without any issues and this is a official copy for the ACM repo

## Merge criteria:
No issues visible after review

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ The main objective is to showcase that a user can take a trained model, use a pi
    ```
    $  oc apply -k acm/registration
    ```
+
    * Core - Cluster host the ODH Core components that will be used in the MLOps Engineer workflow to train, build and push the model.  This cluster is not required to be co-located with the ACM Hub but we group them together to simplify the use case
    * Near Edge - Cluster(s) that will host the running model at the edge.  This is the target environment after a new model is available for use
 1. Deploy Open Data Hub to the Core cluster and register any configurations to support pushing models to the edge cluster
--- a/acm/registration/odh-core/odh-core.application.yaml
+++ b/acm/registration/odh-core/odh-core.application.yaml
@@ -1,0 +1,81 @@
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: odh-core-acm
+---
+apiVersion: app.k8s.io/v1beta1
+kind: Application
+metadata:
+  name: odh-core
+  namespace: odh-core-acm
+spec:
+  componentKinds:
+  - group: apps.open-cluster-management.io
+    kind: Subscription
+  descriptor: {}
+  selector:
+    matchExpressions:
+      - key: app
+        operator: In
+        values: 
+          - odh-core
+---
+apiVersion: apps.open-cluster-management.io/v1
+kind: Channel
+metadata:
+  annotations:
+    apps.open-cluster-management.io/reconcile-rate: medium
+  name: channel
+  namespace: odh-core-acm
+spec:
+  type: Git
+  pathname: 'https://github.com/opendatahub-io/ai-edge'
+---
+apiVersion: cluster.open-cluster-management.io/v1beta2
+kind: ManagedClusterSet
+metadata:
+  name: odh-core
+spec:
+  clusterSelector:
+    selectorType: ExclusiveClusterSetLabel
+---
+apiVersion: cluster.open-cluster-management.io/v1beta2
+kind: ManagedClusterSetBinding
+metadata:
+  name: odh-core
+  namespace: odh-core-acm
+spec:
+  clusterSet: odh-core
+---
+apiVersion: cluster.open-cluster-management.io/v1beta1
+kind: Placement
+metadata:
+  labels:
+    app: odh-core-acm
+  name: placement
+  namespace: odh-core-acm
+spec:
+  clusterSets:
+    - odh-core
+  predicates:
+    - requiredClusterSelector:
+        labelSelector: {}
+---
+apiVersion: apps.open-cluster-management.io/v1
+kind: Subscription
+metadata:
+  annotations:
+    apps.open-cluster-management.io/git-branch: main
+    apps.open-cluster-management.io/git-path: acm/odh-core
+    apps.open-cluster-management.io/reconcile-option: merge
+  labels:
+    app: odh-core
+  name: subscription
+  namespace: odh-core-acm
+spec:
+  channel: github-opendatahub-io-ai-edge-ns/github-opendatahub-io-ai-edge
+  placement:
+    placementRef:
+      kind: Placement
+      name: odh-core-placement
--- a/acm/registration/odh-core/kustomization.yaml
+++ b/acm/registration/odh-core/kustomization.yaml
@@ -2,5 +2,4 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
 resources:
-- odh-core/
-- near-edge/
+- odh-core.application.yaml
--- a/acm/registration/odh-core/kustomization.yaml
+++ b/acm/registration/odh-core/kustomization.yaml
@@ -3,3 +3,37 @@ kind: Kustomization
 
 resources:
 - odh-core.application.yaml
+
+replacements:
+- source:
+    group: apps.open-cluster-management.io
+    kind: Channel
+    fieldPath: metadata.name
+  targets:
+  - select:
+      group: apps.open-cluster-management.io
+      kind: Subscription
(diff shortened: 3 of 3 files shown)


acm/registration/odh-core.application.yaml
```diff
@@ -0,0 +1,50 @@
+apiVersion: app.k8s.io/v1beta1
+kind: Application
+metadata:
+  name: odh-core
+  namespace: rhoai-edge-acm
```
grdryn, line 5: Should there be a Namespace resource before this, to create this namespace?

acm/registration/odh-core.application.yaml
```diff
@@ -0,0 +1,50 @@
...
+    matchExpressions:
+      - key: app
+        operator: In
+        values: 
+          - odh-core
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: ggithubcom-opendatahub-io-ai-edge-ns
```
grdryn, line 21: Very minor: is the extra `g` at the start here (and everywhere else that it's referenced) a typo, or was it intentional?

acm/registration/odh-core.application.yaml
```diff
@@ -0,0 +1,50 @@
...
+  labels:
+    app: odh-core
+  name: odh-core-subscription
+  namespace: rhoai-edge-acm
+spec:
+  channel: ggithubcom-opendatahub-io-ai-edge-ns/ggithubcom-opendatahub-io-ai-edge
+  placement:
+    placementRef:
+      kind: Placement
+      name: odh-core-placement
```
grdryn, line 50: The definition of this Placement resource should also be included here, before the subscription, I think

acm/registration/odh-core.application.yaml
```diff
@@ -0,0 +1,50 @@
...
+kind: Namespace
+metadata:
+  name: ggithubcom-opendatahub-io-ai-edge-ns
+---
+apiVersion: apps.open-cluster-management.io/v1
+kind: Channel
+metadata:
+  annotations:
+    apps.open-cluster-management.io/reconcile-rate: medium
+  name: ggithubcom-opendatahub-io-ai-edge
```
grdryn, line 28: If this channel being in a different namespace is expected to be re-used in other subscriptions, like when we've got an `odh-edge` one, then should it (and the ns definition above it) be repeated in the `odh-edge.yaml`, or should it be pulled out of this file into a common one?

README.md
```diff
@@ -19,8 +19,13 @@ The main objective is to showcase that a user can take a trained model, use a pi
 1. Configure the default Identity Provider
 1. Install Red Hat Advanced Cluster Management
 1. Register the clusters
-   * Core - Cluster host the ODH Core components that will be used in the MLOps Engineer workflow to train, build and push the model
-   * Near Edge - Cluster that will host the running model at the edge.  This is the target environment after a new model is available for use
+   [ACM Application](https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.8/html/applications/managing-applications) manifests are located in [acm/registration](acm/registration) to register and configure the target environments required for the AI at the Edge use cases.  The files ca be applied to the ACM hub cluser manually:
+   ```
+   $ kustomize build acm/registration | oc apply -f -
```
grdryn, line 24: Is the `kustomize` part necessary, when there's no `kustomization.yaml` included here? Actually, it doesn't work without it:

```
$ kustomize build acm/registration | oc apply -f -
Error: unable to find one of 'kustomization.yaml', 'kustomization.yml' or 'Kustomization' in directory '/var/home/gryan/src/github.com/opendatahub-io/ai-edge/acm/registration'
error: no objects passed to apply
```

* Without `kustomization.yaml`, this will work:
```suggestion
   $ oc apply -f acm/registration/
```

* Or if you add a `kustomization.yaml`, this might be simpler, now that `kustomize` is built into `kubectl`/`oc`:
```suggestion
   $ oc apply -k acm/registration/
```

LaVLaS: I forgot to mark this as a draft since the kustomization.yaml was missing from the initial PR.

`oc apply -k <kustomize manifest>` is fine.  I have muscle  memory using `kustomize build` for testing compatibility with different kustomize versions for the new ODH manifests

grdryn:
@LaVLaS Do you think `ManagedClusterSets` and `ManagedClusterSetBindings` and other such resources should also be included here?

LaVLaS:
> @LaVLaS Do you think ManagedClusterSets and ManagedClusterSetBindings and other such resources should also be included here?

I was thinking the same thing. I can't remember if the `ManagedClusterSet` should be documented by us and created by the user during the cluster creation process

acm/registration/odh-core.application.yaml
LaVLaS: Corrected. Since it was copied/pasted directly from the ACM Console UI which creates the NS automatically

grdryn:
> > @LaVLaS Do you think ManagedClusterSets and ManagedClusterSetBindings and other such resources should also be included here?
> 
> I was thinking the same thing. I can't remember if the `ManagedClusterSet` should be documented by us and created by the user during the cluster creation process

Yeah, maybe documenting it is better...I don't know enough about these specific resources or the cluster creation process to know. I was mostly asking while I was trying to figure out how to create the namespace binding on the cluster set, since that's a thing that's needed if we create a new application in a namespace that the cluster set doesn't already have a binding to. So I think at least a `ManagedClusterSetBinding` is what we need to create here when we create a new application namespace, whether we have the `ManagedClusterSet` documented for manual creation or applied from here.

grdryn:
Also, I see this PR is still marked as a draft/WIP. When it's ready to be reviewed, don't forget to remove the draft status :slightly_smiling_face:

grdryn (commented):
It seems like some of the resources refer to the wrong names here. I've suggested a fix for that, using `replacements` in the kustomization.yaml

acm/registration/odh-core/odh-core.application.yaml
```diff
@@ -0,0 +1,73 @@
...
+  type: Git
+  pathname: 'https://github.com/opendatahub-io/ai-edge'
+---
+apiVersion: cluster.open-cluster-management.io/v1beta2
+kind: ManagedClusterSet
+metadata:
+  name: odh-core
+spec:
+  clusterSelector:
+    selectorType: ExclusiveClusterSetLabel
```
grdryn, line 41: I think the ManagedClusterSetBinding will also be needed here:

```suggestion
apiVersion: cluster.open-cluster-management.io/v1beta2
kind: ManagedClusterSet
metadata:
  name: odh-core
spec:
  clusterSelector:
    selectorType: ExclusiveClusterSetLabel
---
apiVersion: cluster.open-cluster-management.io/v1beta2
kind: ManagedClusterSetBinding
metadata:
  name: odh-core
  namespace: odh-core-acm
spec:
  clusterSet: odh-core
```

^with the potential for some `kustomize` replacements akin to the ones I suggested earlier, to make sure that `spec.clusterSet` here refers to the name of the `ManagedClusterSet` above it.

grdryn: Thinking about this a little more, and playing with the equivalent near-edge stuff, I'm starting to think that maybe the `ManagedClusterSet` _shouldn't_ be specified here, just the `ManagedClusterSetBinding`.

Reasoning: the rest of this is the resources to configure an application in ACM, whereas the `ManagedClusterSet` is more the infrastructure that can be common to many applications. This is emphasized by the fact that `ManagedClusterSet` is a cluster-scoped CRD, while all of the rest of these are namespace-scoped.

Keeping the `ManagedClusterSet` here would start to get weird with `kustomize`, if we do things like create multiple applications (which seems like a normal case) based off a common base, and use things like `namePrefix`: we'd end up with multiple `ManagedClusterSets`, where each one has just one application. We'd then have to make sure that the `ManagedClusters` are all labelled correctly for all of them.

In short: `ManagedClusterSet` should be defined alongside `ManagedClusters`, and `ManagedClusterSetBinding` should be here with the application resources. WDYT?

LaVLaS:
Putting this back on draft ...again.  After building a new cluster for a demo, I now think the `odh-core` registration as defined here not should include deployment of the ODH operator or Gitea Operator as those are infrastructure that we expect to be in installed before the PoC workflow.  While they would definitely be included some GitOps configuration, I think that is outside of the GitOps workflow for this PoC

PR #47, opened 2023-08-30
Use ACM app per app to ensure correct NS placement

grdryn:
It seems like all resources get created in a namespace on the target clusters with the same name as the namespace that the subscription on the hub cluster is created in, despite the resources explicitly having a different namespace set.

This change adds new subscriptions such that the two model apps get created in different namespaces on the target clusters.

## How Has This Been Tested?

This change is currently rolled out to the ACM cluster, using `oc apply -k acm/registration`, and the synchronization has been successful, and the resources can be seen created on the target clusters.

## Merge criteria:

### DIFF
--- a/acm/registration/near-edge/base/near-edge.yaml
+++ b/acm/registration/near-edge/base/near-edge.yaml
@@ -2,13 +2,13 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: rhoai-edge-acm
+  name: near-edge-acm-template
 ---
 apiVersion: app.k8s.io/v1beta1
 kind: Application
 metadata:
-  name: near-edge
-  namespace: rhoai-edge-acm
+  name: application
+  namespace: near-edge-acm-template
 spec:
   componentKinds:
   - group: apps.open-cluster-management.io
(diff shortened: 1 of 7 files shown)


StevenTobin (approved):

acm/registration/near-edge/base/near-edge.yaml
```diff
@@ -19,20 +19,15 @@ spec:
...
-metadata:
-  name: ggithubcom-opendatahub-io-ai-edge-ns
+          - near-edge-acm-template
 ---
 apiVersion: apps.open-cluster-management.io/v1
 kind: Channel
 metadata:
   annotations:
     apps.open-cluster-management.io/reconcile-rate: medium
-  name: ggithubcom-opendatahub-io-ai-edge
```
LaVLaS, line 34: Can we rename from a non-generic `channel` change to show where it is pulling the Application from?

grdryn: @LaVLaS the generic `channel` name (and namespace) gets replaced by kustomize, so that it's an application-specific channel, in the same namespace as the subscription:
image

Currently those both point to the same place (this repo), but if I understood correctly yesterday, we want them to eventually point to Gitea, and they could be different repos there (this is what I was thinking when creating more than one channel here, rather than re-using the existing one that you had created manually for the odh-core application.

Would you like the structure or the names that this config defines to be different in some way?

LaVLaS: Thx for the clarification.  The use of `namePrefix` makes perfect since since we are using `kustomize`

PR #50, opened 2023-08-30
Deploy models to edge clusters

grdryn:
## Description
Adds kustomized Deployment, Service, and Route for each model.

## How Has This Been Tested?

I've only used kustomize locally to check the contents, but will deploy to Gitea now

## Merge criteria:

### DIFF
--- a/acm/odh-edge/apps/bike-rental-app/kustomization.yaml
+++ b/acm/odh-edge/apps/bike-rental-app/kustomization.yaml
@@ -5,6 +5,7 @@ resources:
 - ../../base
 
 namespace: bike-rental-app
+namePrefix: bike-rental-app-
 
 commonLabels:
   app: bike-rental-app-1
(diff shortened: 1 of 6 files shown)


acm/odh-edge/apps/bike-rental-app/kustomization.yaml
```diff
@@ -1,7 +1,53 @@
...
+  - select:
+      group: monitoring.coreos.com
+      kind: ServiceMonitor
+    fieldPaths:
+      - spec.selector.matchLabels.app
+
+images:
+- name: edge-model-template-image
+  newName: quay.io/rhoai-edge/bike-rentals-auto-ml
+  digest: sha256:2e973cd30b42ba10cc2cec72570db7e00f8de772fd00c6028ee07b39fc41682f
```
grdryn, line 53: This (but on Gitea) is what I expect a pipeline task to update once there's a new image for this application. Similar for the tensorflow-housing-app one.

LaVLaS: Correct.  For the basic demo, this is really the only value that will change.  As the PoC evolves then we will introduce a growing number of changes as part of the workflow

PR #52, opened 2023-08-30
Expose multiple ports in testing pipeline

piotrpdev:
## Description

Combine multiple `oc expose` statements into one to save time and resources, also expose the gRPC port and label them in the service. (also realized Steven's commit broke the pipeline because of a small detail)

@grdryn Do you want to include this change in your service in `odh-edge`?
@StevenTobin Are you okay with this? This is just the pipeline for testing anyway but it probably doesn't hurt to future-proof in case we add tests for metrics/gRPC.

## How Has This Been Tested?

I tested it using the commands in the pipelines README, on the `local-cluster/azureml-model-to-edge` namespace.

## Merge criteria:

### DIFF
--- a/pipelines/tekton/test-mlflow-image-pipeline/test-mlflow-image-pipeline.yaml
+++ b/pipelines/tekton/test-mlflow-image-pipeline/test-mlflow-image-pipeline.yaml
@@ -71,25 +71,37 @@ spec:
     taskRef:
       kind: ClusterTask
       name: openshift-client
-  - name: create-default-service
+  - name: create-service
     params:
     - name: SCRIPT
-      value: oc expose deployment  $(params.model-name)-$(params.model-version) --port=8080
-        --target-port=8080 --selector='app=$(params.model-name)-$(params.model-version)'
-        --dry-run=client -o yaml |  oc apply -f -
-    - name: VERSION
-      value: latest
-    runAfter:
-    - deploy-container
-    taskRef:
-      kind: ClusterTask
-      name: openshift-client
-  - name: create-metrics-service
-    params:
-    - name: SCRIPT
-      value: oc expose deployment  $(params.model-name)-$(params.model-version) --name=metrics --port=8082
-        --target-port=8082 --selector='app=$(params.model-name)-$(params.model-version)'
-        --dry-run=client -o yaml |  oc apply -f -
+      value: |
+        cat <<EOF | oc apply -f -
+        apiVersion: v1
+        kind: Service
+        metadata:
+          creationTimestamp: null
+          labels:
+            app: $(params.model-name)-$(params.model-version)
+          name: $(params.model-name)-$(params.model-version)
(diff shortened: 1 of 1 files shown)


pipelines/tekton/test-mlflow-image-pipeline/test-mlflow-image-pipeline.yaml
```diff
@@ -71,25 +71,37 @@ spec:
...
-      name: openshift-client
-  - name: create-metrics-service
-    params:
-    - name: SCRIPT
-      value: oc expose deployment  $(params.model-name)-$(params.model-version) --name=metrics --port=8082
-        --target-port=8082 --selector='app=$(params.model-name)-$(params.model-version)'
-        --dry-run=client -o yaml |  oc apply -f -
+      value: |
+        cat <<EOF | oc apply -f -
+        apiVersion: v1
```
LaVLaS, line 79: Can we move this `Service` manifest to a configmap and mount as a Workspace?  That would allow some flexibility in the configuration of the service.

grdryn: I guess the gitops config will need a corresponding update ([here-ish](https://github.com/opendatahub-io/ai-edge/blob/main/acm/odh-edge/apps/bike-rental-app/kustomization.yaml)), right? I wonder if using `kustomize` here with that would be a good option...that way changes wouldn't need to be made both here and there?

piotrpdev: > Can we move this `Service` manifest to a configmap and mount as a Workspace? That would allow some flexibility in the configuration of the service.

I guess I can read the ConfigMap into an env variable in Tekton using [this workaround](https://github.com/tektoncd/pipeline/issues/2688#issuecomment-635365546) and then use `sed` to substitute in the values. I'll try that.

piotrpdev: `error: loading KV pairs: file sources: [../../../acm/odh-edge/base/model-service.yaml]: security; file '/home/razer/ai-edge/acm/odh-edge/base/model-service.yaml' is not in or below '/home/razer/ai-edge/pipelines/tekton/test-mlflow-image-pipeline'` Guess we either need two identical files or a symlink

piotrpdev: Nevermind, the symlink route doesn't work: `error: loading KV pairs: file sources: [model-service.yaml]: security; file '/home/razer/ai-edge/pipelines/tekton/test-mlflow-image-pipeline/model-service.yaml' is not in or below '/home/razer/ai-edge/pipelines/tekton/test-mlflow-image-pipeline'`

Issue #11, opened 2023-08-11
As a MLOps engineer I would like to use git as my source of truth for the deployment of my containerized model, so I can manage its lifecycle using a GitOps approach.


LaVLaS:
Adding requirements that the PoC use case will be using ACM to manage the cluster configurations user GitOps and model lifecycle

Connected User Stories:
- As a MLOps engineer I would like that once my definition of the containerized model is defined in Git, ACM reconciles the desired state with my cluster at the edge. So I have a fully automated deployment mechanism.
- As a MLOps engineer I would also like to manage with ACM the infrastructure needed to deploy models so I can have a single management tool for the whole solution.  

Acceptance Criteria:
- [x] #28
- [ ] Document the GitOps process and how the cluster will be managed
- [ ] Cluster manifests are defined in a GitOps repo
- [ ] Model deployment and configuration for edge deployments are defined in a GitOps repo
- [ ] ACM Hub will apply manifests correctly when initiated by the Edge cluster
- [ ] Document the configuration and lifecycle of the Edge clustes

adelton:
With https://github.com/opendatahub-io/ai-edge/pull/54 and https://github.com/opendatahub-io/ai-edge/pull/137 merged, isn't this actually done?

LaVLaS:
Closing this out as that phase of the PoC has ended

PR #54, opened 2023-08-31
Update image reference in edge Deployment

grdryn:
## Description

Adds another pipeline that changes the image reference in the `acm/odh-edge/apps/.../kustomization.yaml` file, pushes a new branch with the change to the on-prem Git clone of the `opendatahub/ai-edge` repo, and creates a PR with the changes.

## How Has This Been Tested?

Pipelines tested on the `pplaczek-pipeline-dev` and `gryan-pipelines` namespace on `local-cluster`.
Git tested on `http://gitea-ai-edge-rhoai-edge-acm.apps.rhoai-edge-acm-poc.rhaiseng.com/edge-user-1/ai-edge-gitops`.

See [reply below](https://github.com/opendatahub-io/ai-edge/pull/54#issuecomment-1703167501) for example screenshots.

### Testing instructions
If you would like to try this out, the following steps should be runnable in an isolation on the cluster:
- Create a project with `oc new-project <something>`
- Apply the required resources, with `oc apply -k ./pipelines/tekton/gitops-update-pipeline/`
- You can then either create a PipelineRun for the pipeline through the OpenShift Pipelines UI in the cluster (or presumably using `tkn` CLI), or by using one (or both) of the pre-canned manifests, like so:
```
oc create -f ./pipelines/tekton/gitops-update-pipeline/gitops-update-pipelinerun-bike-rentals.yaml
oc create -f ./pipelines/tekton/gitops-update-pipeline/gitops-update-pipelinerun-tensorflow-housing.yaml
```
- You should then see the pipeline running immediately through the web console, and you should see a PR created in the relevant repo at the end. For the bike-rentals one that will be [here](http://gitea-ai-edge-rhoai-edge-acm.apps.rhoai-edge-acm-poc.rhaiseng.com/edge-user-1/ai-edge-gitops/pulls), and for the tensorflow-housing one it will be [here](http://gitea-ai-edge-rhoai-edge-acm.apps.rhoai-edge-acm-poc.rhaiseng.com/edge-user-2/ai-edge-gitops/pulls). 

## Merge criteria:


grdryn:
To run this on a cluster once applied, I've been specifying `VolumeClaimTemplate` for the `git-workspace`, and the `gitea-edge-user-1-basic-auth` Secret as the source for the `git-basic-auth` workspace.

It currently fails on the `commit-image-ref` task, with `/tekton/scripts/script-0-2csnd: line 46: syntax error: unterminated quoted string`. I don't fully understand this, but I _think_ it's the way the `git-cli` task is doing an `eval '<git script param>'` (I don't see the closing `'`, but I might be wrong).

I'm not sure if this approach of trying to re-use existing tasks is the best one...maybe a single generic task from scratch with multiple steps might be another way.

piotrpdev:
I'll look into it

image

piotrpdev:
Got the code working, filed a PR on your fork: https://github.com/grdryn/ai-edge/pull/1

Pics for anybody looking at this in the future (do we have flying cars yet?):

image

image

image

image

pipelines/tekton/gitops-update-pipeline/gitops-update-pipeline.yaml
```diff
@@ -0,0 +1,140 @@
...
+
+        sed -i -e "s/digest: .*/digest: $(params.imageDigest)/" $(params.imageReferenceFilePath)
+
+        if [ -z "$(git status --porcelain)" ]; then
+            echo "Update didn't cause a modification"
+            exit 0
+        fi
+
+        git commit -am "Update image ref from pipeline"
+        git push origin $PR_BRANCH:$PR_BRANCH
```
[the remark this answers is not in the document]
piotrpdev: I'm pretty sure the apostrophe in "Update **_didn't_** cause a modification" was what caused the issue 😩.

piotrpdev:
This PR is related to #11 btw

pipelines/README.md
```diff
@@ -93,6 +93,6 @@ oc create -f test-mlflow-image-pipeline/test-mlflow-image-pipelinerun-tensorflow
 > If using one of the provided `PipelineRun` files, make sure all of the parameters match your Git environment.
 
 ```bash
-oc apply -k gitops-update-pipeline/
-oc create -f gitops-update-pipeline/gitops-update-pipelinerun-tensorflow-housing.yaml
+oc apply -k tekton/gitops-update-pipeline/
+oc create -f tekton/gitops-update-pipeline/gitops-update-pipelinerun-tensorflow-housing.yaml
```
piotrpdev, line 97: You added `tekton/` here but left the others as is? This also contradicts the notice slightly higher in the README that says to `cd` into `tekton` first

pipelines/tekton/gitops-update-pipeline/gitea-edge-user-1-secret.yaml
```diff
@@ -1,11 +1,11 @@
 apiVersion: v1
 kind: Secret
 metadata:
-  name: gitops-git-user-basic-auth
+  name: gitea-edge-user-1
```
piotrpdev, line 4: What if the end user wants to use something other than gitea? Wouldn't keeping it generic not be better/less confusing?

pipelines/tekton/gitops-update-pipeline/gitops-update-git-workspace-bike-rentals-pvc.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
-  name: git-workspace-pvc
+  name: git-workspace-bike-rentals
```
piotrpdev, line 4: Having separate workspaces might be better, but does it matter since the repo's in them get replaced every time a clone happens anyway? The repo's are also cloned to the workspace using their repo's name so there's no chance of collision too.

pipelines/tekton/gitops-update-pipeline/gitea-edge-user-1-secret.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
```
LaVLaS, line 1: Do these `gitea-edge-user-*` secrets need to be checked in to the repo OR should we document them as a required resource for the PipelineRun?

grdryn: That's a good question, and the same goes for the PVCs for the PipelineRun (don't actually need a dedicated PVC even...tekton can generate one per PipelineRun). We currently don't deploy the PipelineRuns by default, but we do provision these Secrets and PVCs. I think maybe we should check them in as examples that are easy to apply for our purposes (and easy to change for others who want to run the demo on their own infra or base something else on it), and document that there are examples that people can change (and that they should consider not checking-in those changes).

WDYT?

pipelines/tekton/gitops-update-pipeline/gitea-edge-user-1-secret.yaml
```diff
@@ -1,11 +1,11 @@
...
-  name: gitops-git-user-basic-auth
+  name: gitea-edge-user-1
 type: Opaque
 stringData:
+  token: b14e9f5de9fabb48a091a56a24140025fa5416ba
   .git-credentials: http://edge-user-1:b14e9f5de9fabb48a091a56a24140025fa5416ba@gitea-ai-edge-rhoai-edge-acm.apps.rhoai-edge-acm-poc.rhaiseng.com
   .gitconfig: |
-    [credential "http://gitea-ai-edge-rhoai-edge-acm.apps.rhoai-edge-acm-poc.rhaiseng.com"]
-      username = edge-user-1
+    [credential]
```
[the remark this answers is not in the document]
grdryn: It appears to work, yes. I'll double check in a clean namespace.

pipelines/tekton/gitops-update-pipeline/gitea-edge-user-1-secret.yaml
grdryn: I prefer the more specific name here, because the point of it is for this Secret to bring the specifics of the user details on the particular git server to the generic pipeline. That is to say, it's more tied to the git server & user than to the PipelineRun. _Hypothetically,_ there might be many such secrets that are for different Git servers & different users on them, all of which might be used with the generic pipeline in different scenarios. In such a case, generic naming of the secrets wouldn't help. WDYT?

pipelines/tekton/gitops-update-pipeline/gitops-update-git-workspace-bike-rentals-pvc.yaml
grdryn: There are a couple of reasons I think that separate workspaces would be better:

* The PVCs have the ReadWriteOnce access mode, which means that only one Pod (or PipelineRun) can use it at a time. This could cause PipelineRuns for the _other_ repo to block unnecessarily while waiting for the current run to complete (if there were many many different variations to the pipeline (i.e. repos), that got run a lot, that might slow things down. I realise it's an unlikely case, but blocking per repo rather than per workspace seems cleaner.
* Even though, as you say, the repos are cloned to the workspace using the repo name, that could still cause a collision in our examples, since both have the same repo name (just in different orgs), right? Granted, I think the access mode on the PVC would prevent that collision, given that only one PipelineRun should be able to mount the PVC anyway

I actually think using Tekton's ability to generate a PVC per PipelineRun might be even better, but maybe a bit wasteful if there's nothing to clean up old pipeline runs older than a certain age or outside the most recent N runs.

WDYT?

pipelines/README.md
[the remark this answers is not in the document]
piotrpdev: I'd personally prefer all of them start with `tekton/` over that and keep the README where it is since the user is not likely to use the other subdirectories often.

I think we need a README in `pipelines/` to tell the user what it's for, and pointing to _another_ README in `pipelines/tekton` feels a little annoying to me e.g. You read the README at the root of the repo, which points you to a README in `pipelines/`, which points you to a README in `pipelines/tekton`.

idk tho ¯\\_(ツ)_/¯

pipelines/tekton/gitops-update-pipeline/gitea-edge-user-1-secret.yaml
piotrpdev: When you put it like that I totally agree, good argument 👍. This may be a little pedantic but would you mind including a note somewhere that the user isn't limited to gitea anyway? I think I would accidentally assume that if it was me.

pipelines/tekton/gitops-update-pipeline/gitops-update-git-workspace-bike-rentals-pvc.yaml
piotrpdev: You're totally right on both points, didn't think of that 👍. 
<br>

> I actually think using Tekton's ability to generate a PVC per PipelineRun might be even better

You mean [like this](
https://tekton.dev/docs/pipelines/workspaces/#volumeclaimtemplate)? I also like that approach.
<br>

> maybe a bit wasteful if there's nothing to clean up old pipeline runs

[OpenShift Pipelines to the rescue](https://docs.openshift.com/container-platform/4.13/cicd/pipelines/customizing-configurations-in-the-tektonconfig-cr.html#annotations-for-automatic-pruning-taskruns-pipelineruns_customizing-configurations-in-the-tektonconfig-cr). There's also a [cron job](https://gist.github.com/ctron/4764c0c4c4ea0b22353f2a23941928ad).

pipelines/README.md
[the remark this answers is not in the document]
grdryn: Updated, thanks.

pipelines/tekton/gitops-update-pipeline/gitea-edge-user-1-secret.yaml
grdryn: Added some words to the README, thanks.

pipelines/tekton/gitops-update-pipeline/gitops-update-git-workspace-bike-rentals-pvc.yaml
grdryn: Updated to use Tekton's PVC generation per run.

pipelines/tekton/gitops-update-pipeline/gitea-edge-user-1-secret.yaml
grdryn: Removed automatic deployment of these secrets, and made it clearer that they're examples.

pipelines/tekton/gitops-update-pipeline/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Kustomization
+apiVersion: kustomize.config.k8s.io/v1beta1
+
+resources:
+  - https://raw.githubusercontent.com/piotrpdev/catalog/patch-1/task/github-open-pr/0.2/github-open-pr.yaml # Change this once the fix is merged
```
LaVLaS, line 5: This references a fix, ss this still required or need to be updated?

grdryn: @piotrpdev I think that's this one, right? https://github.com/tektoncd/catalog/pull/1191

devguyio (approved):
Nice work! I left a small question but other than that it's good!

pipelines/tekton/gitops-update-pipeline/kustomization.yaml
piotrpdev: Looks like the PR for the fix is still being worked on which might take a while ¯\_(ツ)_/¯

pipelines/tekton/gitops-update-pipeline/gitops-update-pipeline.yaml
```diff
@@ -0,0 +1,127 @@
...
+
+            sed -i -e "s/digest: .*/digest: $(params.imageDigest)/" $(params.imageReferenceFilePath)
+
+            if [ -z "$(git status --porcelain)" ]; then
+                echo "Update did not cause a modification"
+                exit 0
+            fi
+
+            git commit -am "Update image ref from pipeline"
+            git push origin $PR_BRANCH:$PR_BRANCH
```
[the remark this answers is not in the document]
piotrpdev: Yes, the task will fail and the pipeline will stop. I have personally had to encounter this at least 20+ times when fixing bugs with this specific script :weary:

grdryn:
I'm going to make a further change to this PR: upon further reflection I will remove the git secrets and document the requirement for them.

piotrpdev:
> @grdryn I'm going to make a further change to this PR: upon further reflection I will remove the git secrets and document the requirement for them.

Surely providing template files wouldn't hurt though? I hate having to create new files and figure out what they need to look like :weary:

pipelines/tekton/gitops-update-pipeline/example-pipelineruns/gitops-update-pipelinerun-bike-rentals.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  labels:
+    tekton.dev/pipeline: gitops-update-pipeline
+  generateName: gitops-update-pipeline-bike-rentals-
+spec:
```
LaVLaS, line 7: I noticed the error below when I deployed this to a fresh cluster.  I didn't have time to debug but there was no issue when creating the tensorflow-housing PipelineRun.  

```
Error from server (BadRequest): error when creating "pipelines/tekton/gitops-update-pipeline/example-pipelineruns/gitops-update-pipelinerun-bike-rentals.yaml": admission webhook "webhook.pipeline.te
kton.dev" denied the request: mutation failed: cannot decode incoming new object: json: unknown field "spec"
```

grdryn: I think I see the problem...it's an indentation error in the volumeclaimtemplate on line 28. Will fix ASAP, thanks

grdryn: This is fixed now, and I've verified that the fix works on the older cluster. Sorry for the noise!

grdryn:
I've updated this with a fix for the issue that @LaVLaS noticed, and just some docs updates around the GitOps secret that I mentioned earlier, and rebased it on the latest state of main. I don't plan on making any further changes to it. If there are no further review comments on it by then, I'll merge it either later tonight or first thing tomorrow morning. Thanks for all the feedback and reviews so far :slightly_smiling_face:

pipelines/tekton/gitops-update-pipeline/kustomization.yaml
LaVLaS: Just adding comment that we should add an TODO issue to track that PR so we can test and update when it's merged

pipelines/tekton/gitops-update-pipeline/gitops-update-pipeline.yaml
```diff
@@ -0,0 +1,127 @@
...
+        name: git-clone
+      workspaces:
+        - name: output
+          workspace: git-workspace
+        - name: basic-auth
+          workspace: git-basic-auth
+    - name: commit-image-ref
+      params:
+        - name: BASE_IMAGE
+          value: cgr.dev/chainguard/git:root-2.39@sha256:7759f87050dd8bacabe61354d75ccd7f864d6b6f8ec42697db7159eccd491139
```
piotrpdev, line 45: Any particular reason for using this image btw?

pipelines/tekton/gitops-update-pipeline/gitops-update-pipeline.yaml
piotrpdev, line 69: Do you think we should bother moving this to a ConfigMap?

pipelines/README.md
```diff
@@ -64,15 +63,54 @@ oc create -f azureml-container-pipeline/azureml-container-pipelinerun-tensorflow
...
 
 ```bash
-oc apply -k test-mlflow-image-pipeline/
-oc create -f test-mlflow-image-pipeline/test-mlflow-image-pipelinerun-tensorflow-housing.yaml
+oc apply -k tekton/gitops-update-pipeline/
+```
+
+#### Run the GitOps pipeline
+
+The pipeline can be run through the OpenShift Pipelines web console, by specifying all of the details for your environment.
```
piotrpdev, line 95: Doesn't this apply to all of the pipelines? Why state it explicitly here?

pipelines/tekton/gitops-update-pipeline/kustomization.yaml
grdryn: Good idea, added an issue here: https://github.com/opendatahub-io/ai-edge/issues/67

pipelines/README.md
```diff
@@ -64,15 +63,54 @@ oc create -f azureml-container-pipeline/azureml-container-pipelinerun-tensorflow
...
+
+Alternatively, the `tekton/gitops-update-pipeline/example-pipelineruns/` contains some examples that can be modified and used.
```
piotrpdev, line 97: The other pipeline directories contain the pipelinerun files directly, not under a subdirectory. Are we planning to enforce this approach across all of the pipelines?

pipelines/README.md
```diff
@@ -64,15 +63,54 @@ oc create -f azureml-container-pipeline/azureml-container-pipelinerun-tensorflow
...
+In these examples, notice that there are template Secret files for the Git credentials that are referenced by different tasks.
+Create an equivalent Secret with appropriate details for your environment, and change the parameter values in the PipelineRun definition to match.
+The examples have been tested using a Gitea server deployed to an OpenShift cluster, but any equivalent Git forge that has the same pull request API (e.g. GitHub) should work too.
```
piotrpdev, line 100: Please add some spacing for readability.
```suggestion
Alternatively, the `tekton/gitops-update-pipeline/example-pipelineruns/` contains some examples that can be modified and used.

In these examples, notice that there are template Secret files for the Git credentials that are referenced by different tasks.

Create an equivalent Secret with appropriate details for your environment, and change the parameter values in the PipelineRun definition to match.

The examples have been tested using a Gitea server deployed to an OpenShift cluster, but any equivalent Git forge that has the same pull request API (e.g. GitHub) should work too.
```

piotrpdev (approved):
LGTM

pipelines/tekton/gitops-update-pipeline/gitops-update-pipeline.yaml
LaVLaS: This is a generic git workflow so I don't have any issues leaving this embedded in the pipeline definition.  There will be some future cleanup of the pipelines but, IMO, I have no major concerns with this implementation right now

pipelines/tekton/gitops-update-pipeline/gitops-update-pipeline.yaml
grdryn: This is just the default image from the git-cli cluster task that comes with OpenShift Pipelines, so I think it got pulled out to a parameter with the same default value when creating the pipeline through the UI. Is there some other image that you want to suggest instead? If so, maybe that should be tackled as a separate issue where it can be tested.

pipelines/README.md
grdryn: I'm not aware of any plans to _enforce_ anything. #62 is the closest thing I know of which plans to address how things are organised here generally. For the case of pipelinerun files, given that they're just optional examples, I think it's worth separating them from the stuff that's always expected to be provisioned.

I'm also not fully convinced that we should even include PipelineRun files in this repo. I would probably have chosen to just document (and demo) running the pipelines through the UI. I'm not sure which approach is better though.

Let me know if you'd like me to make changes to the other Pipelines to make them consistent with this, or if you want to leave it as part of #62, or if you'd like to suggest something else.

pipelines/README.md
grdryn: That will create a paragraph per sentence, where I would consider this all as a single paragraph. If what you're suggesting is that there are too many words in the paragraph, I think I'll just just delete the last sentence, since I don't think it's actually necessary or useful.

pipelines/README.md
grdryn: I'll remove it

grdryn (commented):
Responding to review comments :)

pipelines/README.md
LaVLaS: I don't have an issue including PipelineRuns and I would say they were extremely helpful in bootstrapping the demo cluster by just changing the git server url and a few secrets.  In the long term, it would be better to standardize on a separate PipelineRuns directory that can be combined with some automation to auto-generate the environment parameter values

pipelines/tekton/gitops-update-pipeline/gitops-update-pipeline.yaml
LaVLaS: Is the `BASE_IMAGE` required since the [git-cli task](https://hub.tekton.dev/tekton/task/git-cli) already defines that image as the default value?

LaVLaS (approved):
For the most part, these Pipelines worked with only expected minor changes required.  Any follow-up changes can be dressed in follow-up PRs to improve clarity and reduce redundancy through additional automation to simplify the need to manually edit the same value across multiple files

pipelines/tekton/gitops-update-pipeline/gitops-update-pipeline.yaml
grdryn: Probably not, but I think the pipeline creation UI pulled that out to a corresponding parameter at the pipeline level, with the same default value; so while it _can_ be overridden by a user, it doesn't _need_ to be. Maybe this is useful for cases where there's a CVE in this image, or they want a newer version of Git?

OTOH, I wonder if this is actually unintentionally pinning the version to this digest, even if the underlying git-cli cluster task updates to a new image over time :thinking:

PR #61, opened 2023-09-06
Add a bike-rental app traffic generator and an initial Grafana Inference Health dashboard

### LINKED ISSUE #32: Create a sample grafana dashboard that displays the metrics that we are collecting.

devguyio:
## Description

- Add a CronJob to generate dummy traffic to the bike-rental-app 
- Add an initial Grafana Inference Health dashboard.

Fixes #32 

## How Has This Been Tested?

The resources were tested via the full ACM GitOps workflow.
## Merge criteria:


acm/odh-edge/apps/bike-rental-app/bike-rental-inference-generator-cronjob.yaml
```diff
@@ -0,0 +1,54 @@
...
+  name: bike-rental-inference-generator
+spec:
+  schedule: "*/1 * * * *"  # This will run the job every hour
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+          - name: test-mlflow-container
+            image: quay.io/devguyio/curl-jq
```
grdryn, line 13: Do you have a source repo for this image? It might be useful to have in a comment

pipelines/models/bike-rentals-auto-ml/test_data_generator.py
```diff
@@ -0,0 +1,37 @@
+import pandas as pd
```
grdryn, line 1: Do I understand correctly that this script is not used in the CronJob, and was instead just used to generate the test-data-json from the dataset.csv?

acm/odh-edge/apps/bike-rental-app/bike-rental-inference-generator-cronjob.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: batch/v1
```
grdryn, line 1: If you expect this file to get deployed by default, I think you'll need to add it to the `resources` list in the  `kustomization.yaml` in the same folder

grdryn (approved):
Generally LGTM with a couple of inline comments.

Nice find on the dataset...I wonder if the demo of this PoC could include the DS engineer persona training the model from this dataset? I don't know anything about what would be needed for that, just a thought (I also realise that the model binary that we've got for this at the moment was trained in AzureML rather than using any ODH stuff).

pipelines/models/bike-rentals-auto-ml/test_data_generator.py
devguyio: Correct

acm/odh-edge/apps/bike-rental-app/bike-rental-inference-generator-cronjob.yaml
devguyio: Thanks!, was on my TO DO list indeed. Done.

acm/odh-edge/apps/bike-rental-app/bike-rental-inference-generator-cronjob.yaml
devguyio: I added the Dockerfile of that, which is now called `ab-jq`

acm/odh-core/acm-observability/grafana-dashboard-edge-inference-health.yaml
```diff
@@ -0,0 +1,231 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  annotations:
+    observability.open-cluster-management.io/dashboard-folder: Edge
+  labels:
+    grafana-custom-dashboard: "true"
+  name: grafana-dashboard-edge-inference-health
+data:
+  edge-inference-health.json: |-
```
grdryn, line 10: I guess the contents of this is what's in `acm/odh-core/acm-observability/files/grafana-edge-inference-health.json`, right? If so, then I think it would be cleaner to use Kustomize's [configMapGenerator](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/configmapgenerator/) functionality to put those two pieces together automatically, rather than having to do it manually whenever the dashboard gets changed). You'll see that there's a `secretGenerator` already in the `kustomization.yaml` which does something similar. WDYT?

acm/odh-edge/apps/bike-rental-app/bike-rental-inference-generator-cronjob.yaml
```diff
@@ -0,0 +1,65 @@
...
+kind: CronJob
+metadata:
+  name: bike-rental-inference-generator
+spec:
+  schedule: "*/5 * * * *"  # This will run the job every hour
```
grdryn, line 6:
```suggestion
  schedule: "*/5 * * * *"
```
^This comment has already fallen out of sync

acm/odh-edge/apps/bike-rental-app/bike-rental-inference-generator-cronjob.yaml
```diff
@@ -0,0 +1,65 @@
...
+              value: "1"  # modify this if needed
+            - name: TEST_ENDPOINT
+              value: "invocations"  # modify this
+            volumeMounts:
+            - name: test-data-volume
+              mountPath: /workspace/test-data
+          volumes:
+          - name: test-data-volume
+            configMap:
+              name: bike-rental-test-data  # This assumes you have a ConfigMap named bike-rental-test-data with the JSON data
```
grdryn, line 63: Should kustomize also create this ConfigMap by default, rather than having the assumption in the comment?

Ensuring that this reference is correct with regards to the name of the ConfigMap can also be handled by kustomize `replacements`, similar to how we ensure that a Route refers to the correct Service [here](https://github.com/opendatahub-io/ai-edge/blob/main/acm/odh-edge/apps/bike-rental-app/kustomization.yaml#L39...L48).

grdryn (approved):
Added some new suggestions (as you can see, `kustomize` is my new hammer, and everything looks very nail-shaped :cowboy_hat_face:)

pipelines/models/ab-jq/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM alpine:3.18.3@sha256:7144f7bab3d4c2648d7e59409f15ec52a18006a128c733fcff20d3a4a54ba44a
+
+# `jq` is required to parse and format JSON
+# `apache2-utils` provides ab benchmarking tool
+# `bash` is required to allow stnadardized scripting in bash
+# `coreutils` is required for using functions like `date -d`
+# hadolint ignore=DL3018
+RUN apk update &&       \
+    apk add --no-cache jq apache2-utils bash coreutils &&  \
```
piotrpdev, line 9: Could probably use `ubi9` for this (alpine is fine though):
- `jq` [via appstream repo](https://cdn-ubi.redhat.com/content/public/ubi/dist/ubi9/9/x86_64/appstream/os/Packages/j/)
- `ab` is also provided by `httpd-tools` [via appstream repo](https://cdn-ubi.redhat.com/content/public/ubi/dist/ubi9/9/x86_64/appstream/os/Packages/h/)
- `bash` and `coreutils` preinstalled even in `ubi9/ubi-micro`

piotrpdev (commented):
Mostly LGTM with a few issues, hope none of my comments come across as rude 🙏.

acm/odh-edge/apps/bike-rental-app/bike-rental-inference-generator-cronjob.yaml
```diff
@@ -0,0 +1,65 @@
...
+  name: bike-rental-inference-generator
+spec:
+  schedule: "*/5 * * * *"  # This will run the job every hour
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+          - name: test-mlflow-container
+            image: quay.io/devguyio/ab-jq
```
piotrpdev, line 13: I invited you to the `rhoai-edge` org on Quay, maybe upload the image there?

image

pipelines/models/ab-jq/Dockerfile
gzaronikas: +1 on the UBI usage. IMHO we should always use UBI.

LaVLaS: If there are missing packages in the ubi-image that require additional epel repos then centos base image is a suitable alternative.  We (ODH) had the same discussions where we needed to work around ubi limitations and it some instances it was easier to use centos-stream

acm/odh-edge/apps/bike-rental-app/bike-rental-inference-generator-cronjob.yaml
devguyio: Done

acm/odh-edge/apps/bike-rental-app/bike-rental-inference-generator-cronjob.yaml
devguyio: Done

pipelines/models/ab-jq/Dockerfile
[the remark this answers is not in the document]
devguyio: Moved to the same `pipelines/containerfiles` directory. It's tested by podman.

pipelines/models/bike-rentals-auto-ml/dataset.csv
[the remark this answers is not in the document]
devguyio: Did proper citation under THIRD-PARTY-NOTICES.txt . If we wanna include the training script we can do that in a separate change I believe. Having the dataset file in the same the same dir as the application of the model is a pretty usual thing IMHO and I don't think there's much difference between that and `/data` . It's a pretty small application anyway. Feel free to propose that change separately as well 👍🏽

pipelines/models/ab-jq/Dockerfile
devguyio: Changed to UBI image.

acm/odh-edge/apps/bike-rental-app/bike-rental-inference-generator-cronjob.yaml
devguyio: Done

pipelines/containerfiles/Containerfile.ab-jq
piotrpdev, line 1: Sorry if this is a little pedantic, but does the `Containerfile` need to be under `pipelines/` at all? It's only used by the ACM code as far as I can see so shouldn't it be put somewhere under `acm/`?

devguyio: I think its location is consistent with the rest of the code, since it's not about being under "pipelines" but more about with the rest of the Containerfiles that're related to the model images. 

This image is related to them for now, even if it's just a tooling image. And I think having all of them in a single folder makes sense for now. 

The whole repo structure in general can use some enhancement, as I brought up in #62 , but I don't think this PR should be blocked because of the location of the Container file there, IMHO.

piotrpdev (approved):
LGTM

devguyio:
There's a small bug in the CronJob that I discovered when I was testing, so I'm gonna push the fix then it can be merged.

devguyio:
Big fixed and this PR is now working properly.

PR #63, opened 2023-09-08
Move the ODH Operator subscription to the fast channel for the v2.x release of ODH

### LINKED ISSUE #60: Update the ODH Operator version to v2.x release to support the new install method
The current ACM application deploys the v1.x version (`kfdef`) of the ODH Operator and we need update the `odh-core` acm application to deploy the v2.x operator that supports the new `DataScienceCluster` CRD to deploy ODH components

LaVLaS:
## Description
Updates the ODH Operator subscription to use the `fast` channel for v2.x installations of ODH.  Additionally adds the `DataScienceCluster` installation to deploy ODH in the cluster

Closes #60

## How Has This Been Tested?
Local deployment into my test cluster

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ The main objective is to showcase that a user can take a trained model, use a pi
    * Core - Cluster host the ODH Core components that will be used in the MLOps Engineer workflow to train, build and push the model.  This cluster is not required to be co-located with the ACM Hub but we group them together to simplify the use case
    * Near Edge - Cluster(s) that will host the running model at the edge.  This is the target environment after a new model is available for use
 1. Deploy Open Data Hub to the Core cluster and register any configurations to support pushing models to the edge cluster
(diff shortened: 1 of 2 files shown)


LaVLaS:
I'm not sure how we should handle the installation of `acm/odh-core/apps` since this is dependent on the DataScienceCluster CRD to be installed by OLM in order for it to work.  Given that we will eventually have access to ArgoCD features, we could use [ArgoCD syncwaves](https://argo-cd.readthedocs.io/en/stable/user-guide/sync-waves/) to stagger the install of the operator before the ODH CRD

grdryn (approved):
I haven't tried this out, but it looks reasonable.

@LaVLaS wrt your comment, I don't see an `acm/odh-core/apps/` directory. Is that a current or future concern? I wonder if things in such a location that depend on the DataScienceCluster CRD existing would be okay failing to deploy for a short period of time until the CRD exists?

piotrpdev (approved):
lgtm

PR #107, opened 2023-09-26
The bucket name does not seem to be needed in aws-storage-config.

adelton:
## Description

The bucket name does not seem to be needed in aws-storage-config, it is specified as aws-bucket-name parameter in PipelineRuns.

## How Has This Been Tested?

Tested manually.

## Merge criteria:

### DIFF
--- a/pipelines/tekton/azureml-container-pipeline/aws-env.yaml
+++ b/pipelines/tekton/azureml-container-pipeline/aws-env.yaml
@@ -5,4 +5,4 @@ metadata:
 
 stringData:
   aws-storage-config: |+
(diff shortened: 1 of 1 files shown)


piotrpdev (approved):
If this works feel free to merge, that config comes from this gist by @israel-hdez so I decided not to mess with it just in case: https://gist.github.com/israel-hdez/1848d5a9788140c6f2e42ce316c70d5b#file-containerfile-with-storage-download-L6

For context, `kserve-storage-initializer` uses Boto3
https://github.com/kserve/kserve/blob/master/python/kserve/kserve/storage/storage.py#L156
https://boto3.amazonaws.com/v1/documentation/api/latest/guide/s3-example-download-file.html

grdryn (approved):
Looks reasonable to me, and easily revertible if we discover that it causes an unexpected issue.

Issue #103, opened 2023-09-25
error building at STEP "COPY --chown=ovms:ovms tensorflow-housing/tf2model /models/1": checking on sources under "/workspace/source/tensorflow-housing": copier: stat: "/tensorflow-housing/tf2model": no such file or directory


adelton:
When running
```
oc apply -k tekton/azureml-container-pipeline/
oc create -f tekton/azureml-container-pipeline/azureml-container-pipelinerun-tensorflow-housing.yaml
```
per https://github.com/opendatahub-io/ai-edge/blob/main/pipelines/README.md, the `build-mflow-container` task fails with
```
STEP 1/10: FROM docker.io/openvino/model_server:latest
Trying to pull docker.io/openvino/model_server:latest...
Getting image source signatures
Copying blob sha256:d70d497b7c84ca7c662a2e258df7e8ee65dad98e3c738abf9a718588ff9bdfa7
Copying blob sha256:442c5371227f1795dd8e31d5c52d8818b5aafc27c90f432aac99032262cf9f9c
Copying blob sha256:f7ee72d24b343ea09b822066552c15f591379fcbfa2a2f916f30c2107db35a07
Copying blob sha256:4f4fb700ef54461cfa02571ae0db9a0dc1e0cdb5577484a6d75e68dc38e8acc1
Copying blob sha256:6fe404a37acb87d12d3707fa23f64b729fce30de5111486ae76a97cfe3c0b719
Copying blob sha256:edaedc954fb53f42a7754a6e2d1b57f091bc9b11063bc445c2e325ea448f8f68
Copying blob sha256:6fe404a37acb87d12d3707fa23f64b729fce30de5111486ae76a97cfe3c0b719
Copying blob sha256:f7ee72d24b343ea09b822066552c15f591379fcbfa2a2f916f30c2107db35a07
Copying blob sha256:9df7057f36486d1c5cee80aab92fec45c4e2f24d128f157ecc8504db2c293d90
Copying blob sha256:edaedc954fb53f42a7754a6e2d1b57f091bc9b11063bc445c2e325ea448f8f68
Copying blob sha256:442c5371227f1795dd8e31d5c52d8818b5aafc27c90f432aac99032262cf9f9c
Copying blob sha256:4f4fb700ef54461cfa02571ae0db9a0dc1e0cdb5577484a6d75e68dc38e8acc1
[...]

Issue #91, opened 2023-09-21
Sending this data to GitHub fails


adelton:
The `open-pr` task fails with
```
Sending this data to GitHub: 
{'head': 'pipeline_c6c562e9-386c-4139-be1f-70b1acd5a00e', 'base': 'main', 'title': 'Auto: update image ref', 'body': 'This pull request has been automatically generated by an OpenShift Pipeline in order to update an image ref. Here are the details:\n\n| Key              | Value                                                                     |\n|------------------|---------------------------------------------------------------------------|\n| Pipeline Name    | gitops-update-pipeline                                                  |\n| PipelineRun Name | gitops-update-pipeline-tensorflow-housing-tx96w                                               |\n| PipelinRun UID   | `c6c562e9-386c-4139-be1f-70b1acd5a00e`                                              |\n| File Path        | `acm/odh-edge/apps/tensorflow-housing-app/kustomization.yaml`                                        |\n| New Digest       | `sha256:19e7312d8cbfcbf21dcab083f9ef99af2f807f156b1216a423267010b7fdd7c0`                                                   |\n'}
Error: 401
b'{"message":"user does not exist [uid: 0, name: , keyid: 0]","url":"http://edge-gitea.apps.redacted-1234567.openshiftapps.com/api/swagger"}\n'
```

Neither `README.md` nor `pipelines/README.md` anywhere state that to follow them, GitHub account/repo/something is needed. The `pipelines/README.md` says
> Clone/Mirror this repository on your Git server
placing no further requirements on that server.

The `pipelines/tekton/gitops-update-pipeline/example-pipelineruns/example-git-credentials-secret.yaml` says
```
  # This "token" field is used for interacting with the GitHub API to
  # create a pull request. Change the value to a token from your
  # GitHub-compatible forge.
  token: EXAMPLE_API_TOKEN
```
but is Gitea a GitHub-compatible forge? Given http://gitea-ai-edge-rhoai-edge-acm.apps.rhoai-edge-acm-poc.rhaiseng.com is configured in multiple places, I'd expect the YAMLs to lead the reader through a setup where PRs can be created in the type of Git server that the repo itself showcases, or that creating the PRs would be optional and would not break the whole flow.

grdryn:
This appears to work fine when I try it. A PR in Gitea gets created as expected.

It might be possible that you had a typo on the access token that you created and added to the secret? Or is part of what you're saying that you didn't update the secret because there's a gap in documentation?

[...]

PR #30, opened 2023-08-25
Add manifest files for deploying the gitea-operator

LaVLaS:
## Description
Adds support for deploying a Gitea server in the cluster to enable a target for the gitops workflow

Closes #28

## How Has This Been Tested?
Deployed and testing the in demo cluster

## Merge criteria:
Manifest successfully deploys the gitea-operator which also deploy a gitea-server application

PR #108, opened 2023-09-26
For the container build, just use the upstream repo.

adelton:
## Description

For the container build, just use the upstream repo.

For the GitOps later, we might prescribe to use a clone. But for now it's better to point to the up-to-date upstream than the stale gitea content.

Fixes https://github.com/opendatahub-io/ai-edge/issues/103.

## How Has This Been Tested?

Manually.

## Merge criteria:

### DIFF
--- a/pipelines/tekton/azureml-container-pipeline/azureml-container-pipelinerun-bike-rentals.yaml
+++ b/pipelines/tekton/azureml-container-pipeline/azureml-container-pipelinerun-bike-rentals.yaml
@@ -13,11 +13,11 @@ spec:
   - name: aws-bucket-name
     value: rhoai-edge-models
   - name: gitServer
-    value: http://gitea-ai-edge-rhoai-edge-acm.apps.rhoai-edge-acm-poc.rhaiseng.com
+    value: https://github.com
   - name: gitOrgName
-    value: edge-user-1
+    value: opendatahub-io
   - name: gitRepoName
-    value: ai-edge-gitops
+    value: ai-edge
   - name: containerfileRelativePath
     value: pipelines/containerfiles/Containerfile.seldonio.mlserver.mlflow
   pipelineRef:
--- a/pipelines/tekton/azureml-container-pipeline/azureml-container-pipelinerun-tensorflow-housing.yaml
+++ b/pipelines/tekton/azureml-container-pipeline/azureml-container-pipelinerun-tensorflow-housing.yaml
@@ -13,11 +13,11 @@ spec:
   - name: aws-bucket-name
     value: rhoai-edge-models
   - name: gitServer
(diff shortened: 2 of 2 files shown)


piotrpdev:
@grdryn @LaVLaS any opinion on this?

grdryn (commented):
I don't have strong opinions about this, but if we go this way then I wonder if there's use in having the Gitea server at all?

I wonder if we should alternatively just remember to update the Gitea repos when we merge changes that affect them? IIRC this was part of the original expectations of the PoC, so I wonder if it should be kept unless those expectations have changed?

adelton:
Will we be able to make the GitOps PR work with Gitea -- https://github.com/opendatahub-io/ai-edge/issues/91?

grdryn:
> Will we be able to make the GitOps PR workflows work with Gitea -- #91?

It worked fine a couple of weeks ago :thinking: I hadn't seen that issue yet, will take a look now, thanks.

LaVLaS (approved):
I 100% agree with this since this repo is the source of truth.  We can follow up with another issue to move these environment specific configurations to an external (to the pipeline) config file that can be mounted in workspace

grdryn:
> I 100% agree with this since this repo is the source of truth. We can follow up with another issue to move these environment specific configurations to an external (to the pipeline) config file that can be mounted in workspace

@LaVLaS since you're the person who has demoed this most often, your opinion carries the most weight :)

This is just the parameters for the example `PipelineRun`s anyway, so anyone should be able to modify them if they want to run targeting a different Git server. In that sense, it's already "external config" right?

grdryn:
So, should we merge this change so that image builds use the GitHub repo; and then keep the GitOps repo using Gitea, or should that be changed too?

adelton:
Which GitOps repo do you have in mind?

adelton:
Rebased on latest main to make the review easier.

grdryn:
@adelton my question was probably caused by confusion on my part. I originally thought that this PR was around getting rid of Gitea altogether out of the PoC, but I see now that it's just for the containerfile in the image build, and the pipeline to update the resources with the new image reference in the Gitea repo is staying the same.

If that understanding is correct, then this looks good to me.

grdryn (approved):

adelton:
Right, currently I just try to make things work.

We might want to have steps about using branch of the GitHub repo, or having completely separate git server, possibly setting up a Gitea on the primary cluster, but that would be later changes.

grdryn:
> We might want to have steps about using branch of the GitHub repo, or having completely separate git server, possibly setting up a Gitea on the primary cluster, but that would be later changes.

Deploying Gitea is a work in progress by Landon, in #30

grdryn:
Given that there aren't any unresolved concerns about this, I'll merge it now. Thanks!

Issue #56, opened 2023-09-05
Containerfile for image build should come from Git rather than S3


grdryn:
In the near-edge PoC, the image build pipeline downloads the model file(s) from AWS S3. It also downloads the Containerfile from the S3 bucket. It would be better if the Containerfile was stored in a Git repository. The reason the model is stored in S3 is that it's a pre-built binary artifact, and wouldn't be very suitable for storage in Git if it's a large model file. The Containerfile is source code though, so it should be stored in Git as normal.

I believe that this will require changing the image build pipeline to clone the repo that contains the containerfile, since I don't believe it does a Git clone of anything right now.

piotrpdev:
I talked about why I took this approach for now [in the Pull Request with these changes](https://github.com/opendatahub-io/ai-edge/pull/38#discussion_r1309331077). Here it is:

> Yes, I thought about using a symlink or something but haven't used them in a Git context before. I thought that providing the end user a template Containerfile they can include with their model and modify to include additional details would be useful e.g. extra flags, packages, etc.
>
> A cleaner approach would be to use something like [Build Time Variables](https://docs.docker.com/engine/reference/builder/#arg) and provide details to a template Containerfile at build time using tekton e.g. the name of the model, additional modelserver flags etc. This way the user wouldn't have to include the Containerfile in their model files. However, I have tried and not been able to get these to work very well, especially when used for `CMD` or `ENTRYPOINT` statements [[1]](https://docs.docker.com/engine/reference/builder/#environment-replacement) [[2]](https://github.com/moby/moby/issues/18492)
>
> Another approach could involve using a generic name for the model and paths, in this scenario we could again use a template Containerfile and tekton. However, the model name is included in the endpoints, and further issues could arise e.g. we don't know what kind of behavior would result from deploying multiples instances of a model with the same generic name, the logs could be hard to distinguish from other model logs, etc.

piotrpdev:
I think I'm making some progress on this, managed to find a workaround in the `Containerfile`s that should hopefully be stable.

piotrpdev:
@grdryn @LaVLaS would you prefer the `Containerfile` to be download from Git or stored in a ConfigMap? They're generic templates that are meant to be used with many models anyway and I think I've gotten Build Args to work so the model name, port, etc. could be passed in through the pipeline.

grdryn:
[...]

PR #112, opened 2023-10-02
[RHOAIEDGE-15] Adding option to fetch from git

### LINKED ISSUE #57: Consider fetching one model from Git rather than S3
We have two example models, and associated image build pipelines that pull those models from S3. As a reusable demo, it would be easier to set up if one of these (the smaller one, ideally) was pulled from Git, so that the person demoing can just interact with that one, and doesn't have to also have an AWS account and access to an S3 bucket (either a shared one that we provide the location for, or one of their own where they can put a trained model).

biswassri:
## Description

Fixes #57 
Added a git-clone task to clone the model repo from a git source. This addition required some model paths to be updated in multiple places.

## How Has This Been Tested?

Created a new pipeline run with the changes in my `srbiswas-pipeline-dev` project  and was able to have a successful pipeline run.

 

## Merge criteria:

### DIFF
--- a/pipelines/tekton/build-container-image-pipeline/build-container-image-pipeline.yaml
+++ b/pipelines/tekton/build-container-image-pipeline/build-container-image-pipeline.yaml
@@ -25,8 +25,10 @@ spec:
     type: string
   - name: git-model-repo
     type: string
+    default: ""
   - name: git-revision
     type: string
+    default: "main"
   tasks:
   - name: git-clone-model-repo
     taskRef:
--- a/pipelines/tekton/build-container-image-pipeline/build-container-image-pipeline.yaml
+++ b/pipelines/tekton/build-container-image-pipeline/build-container-image-pipeline.yaml
@@ -78,6 +78,9 @@ spec:
       value: registry.redhat.io/openshift-pipelines/pipelines-git-init-rhel8@sha256:1a50511583fc02a27012d17d942e247813404104ddd282d7e26f99765174392c
     - name: subdirectory
       value: /containerfile_repo/
+    runAfter:
+    - kserve-download-model
+    - git-clone-model-repo
     taskRef:
       kind: ClusterTask
       name: git-clone
@@ -86,9 +89,6 @@ spec:
       workspace: buildah-cache
     - name: basic-auth
       workspace: git-basic-auth
-    runAfter:
-    - kserve-download-model
-    - git-clone-model-repo
   - name: check-model-and-containerfile-exists
     params:
     - name: model-name
--- a/pipelines/tekton/build-container-image-pipeline/build-container-image-pipeline.yaml
+++ b/pipelines/tekton/build-container-image-pipeline/build-container-image-pipeline.yaml
@@ -21,9 +21,7 @@ spec:
     type: string
   - name: modelRelativePath
     type: string
-  - name: fetch-git-model
-    type: string
-  - name: fetch-s3-model
+  - name: fetch-model
     type: string
   - name: git-model-repo
     type: string
@@ -36,7 +34,7 @@ spec:
       name: git-clone
     params:
     - name: url
-      value: $(params.git-model-repo) 
+      value: $(params.git-model-repo)
     - name: revision
       value: $(params.git-revision)
     - name: gitInitImage
@@ -47,9 +45,9 @@ spec:
     - name: output
       workspace: buildah-cache
     when:
-    - input: "$(params.fetch-git-model)"
+    - input: "$(params.fetch-model)"
       operator: in
-      values: ["true"]
+      values: ["git"]
   - name: kserve-download-model
     params:
     - name: model-name
@@ -67,25 +65,28 @@ spec:
     - name: aws-secret
       workspace: aws-secret
     when:
-    - input: "$(params.fetch-s3-model)"
+    - input: "$(params.fetch-model)"
       operator: in
-      values: ["true"]
+      values: ["s3"]
   - name: git-clone-containerfile-repo
     params:
-      - name: url
-        value: $(params.gitServer)/$(params.gitOrgName)/$(params.gitRepoName)
-      - name: gitInitImage
-        value: registry.redhat.io/openshift-pipelines/pipelines-git-init-rhel8@sha256:1a50511583fc02a27012d17d942e247813404104ddd282d7e26f99765174392c
-      - name: subdirectory
-        value: /containerfile_repo/
+    - name: url
+      value: $(params.gitServer)/$(params.gitOrgName)/$(params.gitRepoName)
+    - name: gitInitImage
+      value: registry.redhat.io/openshift-pipelines/pipelines-git-init-rhel8@sha256:1a50511583fc02a27012d17d942e247813404104ddd282d7e26f99765174392c
+    - name: subdirectory
+      value: /containerfile_repo/
     taskRef:
       kind: ClusterTask
       name: git-clone
     workspaces:
-      - name: output
-        workspace: buildah-cache
-      - name: basic-auth
-        workspace: git-basic-auth
+    - name: output
+      workspace: buildah-cache
+    - name: basic-auth
+      workspace: git-basic-auth
+    runAfter:
+    - kserve-download-model
+    - git-clone-model-repo
   - name: check-model-and-containerfile-exists
     params:
     - name: model-name
@@ -97,7 +98,7 @@ spec:
     - name: current-namespace
       value: $(context.pipelineRun.namespace)
     runAfter:
-      - git-clone-containerfile-repo
+    - git-clone-containerfile-repo
     taskRef:
       kind: Task
       name: check-model-and-containerfile-exists
@@ -112,7 +113,7 @@ spec:
     - name: VERSION
       value: latest
     runAfter:
-     - check-model-and-containerfile-exists
+    - check-model-and-containerfile-exists
     taskRef:
       kind: ClusterTask
       name: openshift-client
--- a/pipelines/tekton/build-container-image-pipeline/build-container-image-pipeline.yaml
+++ b/pipelines/tekton/build-container-image-pipeline/build-container-image-pipeline.yaml
@@ -19,7 +19,37 @@ spec:
     type: string
   - name: containerfileRelativePath
     type: string
+  - name: modelRelativePath
+    type: string
+  - name: fetch-git-model
+    type: string
+  - name: fetch-s3-model
+    type: string
+  - name: git-model-repo
+    type: string
+  - name: git-revision
+    type: string
   tasks:
+  - name: git-clone-model-repo
+    taskRef:
+      kind: ClusterTask
+      name: git-clone
+    params:
+    - name: url
+      value: $(params.git-model-repo) 
+    - name: revision
+      value: $(params.git-revision)
+    - name: gitInitImage
+      value: registry.redhat.io/openshift-pipelines/pipelines-git-init-rhel8@sha256:1a50511583fc02a27012d17d942e247813404104ddd282d7e26f99765174392c
+    - name: subdirectory
+      value: /model_dir/
+    workspaces:
+    - name: output
+      workspace: buildah-cache
+    when:
+    - input: "$(params.fetch-git-model)"
+      operator: in
+      values: ["true"]
   - name: kserve-download-model
     params:
     - name: model-name
@@ -36,6 +66,10 @@ spec:
       workspace: buildah-cache
     - name: aws-secret
       workspace: aws-secret
+    when:
+    - input: "$(params.fetch-s3-model)"
+      operator: in
+      values: ["true"]
   - name: git-clone-containerfile-repo
     params:
       - name: url
@@ -44,8 +78,6 @@ spec:
         value: registry.redhat.io/openshift-pipelines/pipelines-git-init-rhel8@sha256:1a50511583fc02a27012d17d942e247813404104ddd282d7e26f99765174392c
       - name: subdirectory
         value: /containerfile_repo/
-    runAfter:
-      - kserve-download-model
     taskRef:
       kind: ClusterTask
       name: git-clone
@@ -58,6 +90,8 @@ spec:
     params:
     - name: model-name
       value: $(params.model-name)
+    - name: modelRelativePath
+      value: model_dir/$(params.modelRelativePath)
     - name: containerfilePath
       value: containerfile_repo/$(params.containerfileRelativePath)
     - name: current-namespace
@@ -93,7 +127,7 @@ spec:
     - name: DOCKERFILE
       value: containerfile_repo/$(params.containerfileRelativePath)
     - name: CONTEXT
-      value: $(params.model-name)
+      value: model_dir/$(params.modelRelativePath)/$(params.model-name)
     - name: TLSVERIFY
       value: "true"
     - name: FORMAT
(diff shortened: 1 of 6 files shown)


biswassri:
Other options that I've tried: 
1. `adding curl in one step` : but the base image doesn't contain curl in that case 
2. `using when condition`:  This field is not supported in the step level in Tekton

Problems with the above changes: 
1. Currently passing the git path through a param but the download is dependent on the extension of the file (here it's .pkl). Should we parameterize the file extension also? Is that a good solution? 

2. The step that is false, still has a container created. Are we good with that approach? 

I'm fairly new to tekton so all suggestions are welcome :)

pipelines/tekton/azureml-container-pipeline/kserve-download-model.yaml
```diff
@@ -12,13 +12,30 @@ spec:
...
+    type: string
+    default: https://github.com/opendatahub-io/ai-edge/raw/3dd0242871371cf66af3f20e3a8924a4b5fc1658/pipelines/models/bike-rentals-auto-ml/model.pkl
   steps:
-  - name: download-model
+  - name: download-model-github
+    image: curlimages/curl:latest
+    script: |
+      if [ "$(params.fetch-github-model)" = "true" ]; then
+        curl -L $(params.github-url) -o $(workspaces.workspace.path)/$(params.model-name).pkl
+      fi
```
piotrpdev, line 30: I believe git clone is preferable over curl in our use case: https://github.com/opendatahub-io/ai-edge/issues/56#issuecomment-1717325175

pipelines/tekton/azureml-container-pipeline/kserve-download-model.yaml
```diff
@@ -12,13 +12,30 @@ spec:
...
     type: string
+  - name: fetch-github-model
+    type: string
+    default: "true"
+  - name: fetch-S3-model
+    type: string
+    default: "false"
+  - name: github-url
+    type: string
+    default: https://github.com/opendatahub-io/ai-edge/raw/3dd0242871371cf66af3f20e3a8924a4b5fc1658/pipelines/models/bike-rentals-auto-ml/model.pkl
```
piotrpdev, line 23: The image build process for the `bike-rentals-auto-ml` model needs more than just the `model.pkl` file, it includes the dependencies in `conda.yaml` for example. You should download the entire folder for the model.

pipelines/tekton/azureml-container-pipeline/kserve-download-model.yaml
```diff
@@ -12,13 +12,30 @@ spec:
     type: string
   - name: aws-bucket-name
     type: string
+  - name: fetch-github-model
+    type: string
+    default: "true"
+  - name: fetch-S3-model
+    type: string
+    default: "false"
```
piotrpdev, line 20: We're planning on having more options e.g. git, s3, azure, pvc, etc. So it would probably be better to just replace these with a single parameter e.g. "download-location" and use `if [ "$(params.fetch-github-model)" = "S3" ];` etc.

pipelines/tekton/azureml-container-pipeline/kserve-download-model.yaml
biswassri: That's right, I did try moving it to the pipelineRun file, but it cause the pipelineRun creation to break. Now that you suggested I move to gitclone, I'll try moving it up again.

pipelines/tekton/azureml-container-pipeline/kserve-download-model.yaml
biswassri: That's a good catch, I completely forgot that it's just not these two.

piotrpdev: You're probably right, I like this approach more too.

LaVLaS: As this PoC is evolving, this can definitely be an issue to follow-up later as we start discussion on a standard MLOps pipeline

pipelines/tekton/azureml-container-pipeline/kserve-download-model.yaml
```diff
@@ -12,13 +12,30 @@ spec:
...
+  - name: github-url
+    type: string
+    default: https://github.com/opendatahub-io/ai-edge/raw/3dd0242871371cf66af3f20e3a8924a4b5fc1658/pipelines/models/bike-rentals-auto-ml/model.pkl
   steps:
-  - name: download-model
+  - name: download-model-github
+    image: curlimages/curl:latest
```
adelton, line 26: Please fully qualify the image name.

pipelines/tekton/azureml-container-pipeline/kserve-download-model.yaml
adelton: > Instead of overloading this task should we be moving this to dedicated Tasks for kserve-download-model-s3, kserve-download-model-git, ...kserve-download-model-* and ...

I agree with having separate Tasks but I wonder about the `kserve-*` naming.

What kserve-ish is about git-clone or in general fetching the model? I understand we now use `quay.io/opendatahub/kserve-storage-initializer:v0.11` which comes from https://github.com/opendatahub-io/kserve which is forked from https://github.com/kserve/kserve ... but is we talk about fetching the model in build-time, what role does kserve play?

biswassri: > > Instead of overloading this task should we be moving this to dedicated Tasks for kserve-download-model-s3, kserve-download-model-git, ...kserve-download-model-* and ...
> 
> I agree with having separate Tasks but I wonder about the `kserve-*` naming.
> 
> What kserve-ish is about git-clone or in general fetching the model? I understand we now use `quay.io/opendatahub/kserve-storage-initializer:v0.11` which comes from https://github.com/opendatahub-io/kserve which is forked from https://github.com/kserve/kserve ... but is we talk about fetching the model in build-time, what role does kserve play?

I agree too, maybe we can move to just model-download-* for individual tasks? I think it would be fine to rename the kserve download model too and won't break anything if it refactor in all places?

pipelines/tekton/azureml-container-pipeline/kserve-download-model.yaml
biswassri: I think we might no longer need this, but if this remains, I'll update with the full image. 👍🏻

pipelines/tekton/azureml-container-pipeline/kserve-download-model.yaml
piotrpdev: > I agree too, maybe we can move to just model-download-* for individual tasks? I think it would be fine to rename the kserve download model too and won't break anything if it refactor in all places?

Sure go ahead, @jackdelahunt will have to resolve merge conflicts though 👀

pipelines/tekton/azureml-container-pipeline/kserve-download-model.yaml
biswassri: @piotrpdev While using git clone, should we be git cloning https://github.com/opendatahub-io/ai-edge.git  or the mlflow repo? Only problem is that it would be very heavy and be cloning other models also. What do you think?

I was back to considering curling the whole folder just.

piotrpdev: Responded to this in DM's TL;DR:
* Clone https://github.com/opendatahub-io/ai-edge.git because that's where the generated model files are.
* Our repo isn't that heavy, make a shallow clone: https://stackoverflow.com/a/34657781/19020549
* Also see https://github.com/opendatahub-io/ai-edge/issues/56#issuecomment-1716393051

pipelines/tekton/azureml-container-pipeline/kserve-download-model.yaml
jackdelahunt: So far I have been testing my stuff on a demo pipeline so it should be fine @biswassri  👍  just make sure to push them so I can add to them

piotrpdev: @biswassri @jackdelahunt #131 got merged so now both of you have to rebase 😵 good luck 👍

LaVLaS:
@biswassri Do these new steps need to be ordered within the main pipeline using [runAfter](https://tekton.dev/docs/pipelines/pipelines/#using-the-runafter-field) ?

biswassri:
> @biswassri Do these new steps need to be ordered within the main pipeline using [runAfter](https://tekton.dev/docs/pipelines/pipelines/#using-the-runafter-field) ?

Actually not. We only have one step(as task) added in the main-pipeline and since it's when condition based it's difficult to be order. So the tasks following the git-clone one are ordered as before. The success of this git-clone task can be confirmed by the `check-model-and-containerfile-exists` hence i've skipped the ordering.

pipelines/tekton/build-container-image-pipeline/build-container-image-pipelinerun-bike-rentals.yaml
```diff
@@ -20,10 +20,20 @@ spec:
     value: ai-edge
   - name: containerfileRelativePath
     value: pipelines/containerfiles/Containerfile.seldonio.mlserver.mlflow
+  - name: fetch-git-model
```
adelton, line 23: If you are changing the default behaviour of the PoC defaults to use the model from git and not from S3 (nothing against that), the `pipelines/README.md` needs to be updated to mark the steps about S3 optional.

Even if S3 stayed as the default, information about the optional approach with the git repo should be added to README as well.

adelton (changes requested):
I've finished the review, some additional changes are likely needed.

pipelines/tekton/build-container-image-pipeline/build-container-image-pipeline.yaml
```diff
@@ -19,7 +19,37 @@ spec:
     type: string
   - name: containerfileRelativePath
     type: string
+  - name: modelRelativePath
```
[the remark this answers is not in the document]
biswassri: Oh yes, since I was only testing out bike-rental pipeline I missed this. I'll update the changes.

pipelines/tekton/build-container-image-pipeline/build-container-image-pipeline.yaml
```diff
@@ -44,8 +78,6 @@ spec:
         value: registry.redhat.io/openshift-pipelines/pipelines-git-init-rhel8@sha256:1a50511583fc02a27012d17d942e247813404104ddd282d7e26f99765174392c
       - name: subdirectory
         value: /containerfile_repo/
-    runAfter:
-      - kserve-download-model
```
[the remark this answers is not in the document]
biswassri: IMO, if `git-clone-model-repo` or `kserve-download-model` doesn't run successfully in before `check-model-and-containerfile-exists` then `check-model-and-containerfile-exists` fails and the following task `create-imagestream` doesn't run at all and the pipeline fails. 
The problem with having `runAfter` in both the dependencies is that it doesn't have a conditions field meaning 
 it doesn't support something like runafter: `kserver-download-model` || `git-clone-model-repo`. Let me know if you'd like to discuss this.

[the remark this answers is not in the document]
adelton: Why can't you just liste them both
```
    runAfter:
      - kserve-download-model
      - git-clone-model-repo
```
?

The `runAfter` is about ordering -- the fact that one of them will always be skipped so it will be a logical `or` does not matter:
Screenshot_2023-11-03_16-26-21

pipelines/tekton/build-container-image-pipeline/build-container-image-pipelinerun-bike-rentals.yaml
adelton: You might actually want to do it in this PR -- that way people reviewing it can cross check if the behaviour change (code / YAML) matches the intended change (the desription of the workflow in the docs).

biswassri: Sure! I'll add it to this PR. My only concern point for that was it would become too many changes in one merge but the corresponding changes with the readme update makes sense.

pipelines/tekton/build-container-image-pipeline/build-container-image-pipeline.yaml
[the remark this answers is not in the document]
biswassri: I probably mis-understood the runAfter mechanism, that it requires each task below it to successfully run. I'll try this  out.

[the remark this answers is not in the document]
adelton: It actually used to be that way so if you knew Tekton before 2021, that's the previous behaviour. It was changed via https://github.com/tektoncd/community/blob/main/teps/0059-skipping-strategies.md, it seems.

pipelines/tekton/build-container-image-pipeline/build-container-image-pipeline.yaml
```diff
@@ -19,7 +19,37 @@ spec:
...
+  - name: fetch-s3-model
+    type: string
+  - name: git-model-repo
+    type: string
+  - name: git-revision
+    type: string
   tasks:
+  - name: git-clone-model-repo
+    taskRef:
+      kind: ClusterTask
```
jackdelahunt, line 35: `ClusterTask` is apparently deprecated 

https://tekton.dev/docs/pipelines/tasks/#task-vs-clustertask

LaVLaS:
Updating from an offline conversation: @biswassri will be updating the affected Pipelineruns with the missing parameters

pipelines/tekton/build-container-image-pipeline/build-container-image-pipeline.yaml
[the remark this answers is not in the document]
biswassri: Fixed this.

pipelines/tekton/build-container-image-pipeline/build-container-image-pipeline.yaml
```diff
@@ -19,7 +19,37 @@ spec:
...
   - name: containerfileRelativePath
     type: string
+  - name: modelRelativePath
+    type: string
+  - name: fetch-model
+    type: string
+  - name: git-model-repo
+    type: string
+    default: ""
+  - name: git-revision
```
jackdelahunt, line 29: Some parameter names are `camelCase` while others are `kebab-case` should we try make these consistent?

jackdelahunt (approved):
LGTM 👍

LaVLaS:
/approve

The build-container workflow  worked for me when deploying locally and fetching from S3.

pipelines/tekton/build-container-image-pipeline/build-container-image-pipeline.yaml
```diff
@@ -19,7 +19,37 @@ spec:
...
+      name: git-clone
+    params:
+    - name: url
+      value: $(params.git-model-repo)
+    - name: revision
+      value: $(params.git-revision)
+    - name: gitInitImage
+      value: registry.redhat.io/openshift-pipelines/pipelines-git-init-rhel8@sha256:1a50511583fc02a27012d17d942e247813404104ddd282d7e26f99765174392c
+    - name: subdirectory
+      value: /model_dir/
```
adelton, line 45: Something about the static `model_dir` path is  wrong: https://github.com/opendatahub-io/ai-edge/issues/175.

Issue #110, opened 2023-09-26
There is a Makefile but no documentation what it's for


adelton:
The repository contains https://github.com/opendatahub-io/ai-edge/blob/main/Makefile but checking all the documentation and grepping the whole repo for `make`, I couldn't find any explanation about its role in the whole setup / PoC.

grdryn:
There's some documentation about newer tasks in the DEVELOPMENT.md now.

There are some undocumented tasks around observability installation still though. @StevenTobin would know the most about those. If we're happy with the installing user using `make`, then they should be documented as-is in the README.md; if not, then the steps they contain could be documented in the README.md.

Issue #87, opened 2023-09-21
Let pipelines use model from PVC


piotrpdev:
## Description

You could provide a local model back in #18 (https://github.com/piotrpdev/azureml-model-to-edge/commit/626d5a5c85ff8c5cdd4019714dce91fe1c1b7751) but that was replaced for the S3 method (and also Git in #57 soon). Ideally a single task should handle getting the model and take some input for how e.g. using Git, S3, Azure, PVC (local), etc. Modifying our existing [`kserve-download-model`](https://github.com/opendatahub-io/ai-edge/blob/e0086c9e2c223c8d4a1df7a415d8aee736365e75/pipelines/tekton/azureml-container-pipeline/kserve-download-model.yaml) Tekton Task is probably best for this.

Since no downloading needs to actually be done for the PVC (local) scenario, you could maybe just skip the `kserve-download-model` Task all together using a [Tekton `when` guard](https://tekton.dev/docs/pipelines/pipelines/#guard-task-execution-using-when-expressions). However, I think modifying the Task to move the model from the provided PVC to the [`buildah-cache`](https://github.com/opendatahub-io/ai-edge/blob/e0086c9e2c223c8d4a1df7a415d8aee736365e75/pipelines/tekton/azureml-container-pipeline/azureml-container-pipeline.yaml#L72) workspace is better, especially since we're probably going to make each `PipelineRun` use [its own PVC soon](https://github.com/opendatahub-io/ai-edge/pull/54#discussion_r1319746380) (using [`volumeClaimTemplate`](https://tekton.dev/docs/pipelines/workspaces/#volumeclaimtemplate)).

<br>

> _Isn't downloading from Git enough?_

The PoC phase two requirements specify using a model stored in a PVC as a _'P0: must have'_ [[1]](https://docs.google.com/document/d/1C47Jw4EC8KNvi01ZKG0Rd8spPBJ0nIx-FbYal20EApc/edit#heading=h.e7hnzf726tke). When RHODS integration and the new Model Registry [[2]](https://docs.google.com/document/d/1G-pjdGaS2kLELsB5kYk_D4AmH-fTfnCnJOhJ8xENjx0/edit?usp=sharing) comes into play this will probably be an important feature.

## A/C

* Provide template PVC file which uses best practices
* Easy and documented way of uploading a local model to the PVC e.g. using a ubi9 pod and `oc cp`
[...]

PR #115, opened 2023-10-03
Draft: Local model upload to PVC

jackdelahunt:
## Merge criteria:


pipelines/tekton/azureml-container-pipeline/model-upload/Makefile
```diff
@@ -0,0 +1,9 @@
+create_pod:
+	cat basic-pod.yaml | sed 's/PVC_NAME/${PVC_NAME}/g' | oc create -f -
+	oc get pod -w basic-pod
+
+upload_model:
+	oc cp ${MODEL_PATH} basic-pod:/workspace/${PVC_NAME}/
+
+clean:
+	cat basic-pod.yaml | sed 's/PVC_NAME/${PVC_NAME}/g' | oc delete -f -
```
piotrpdev, line 9: _"All of the targets here should be declared as `.PHONY` targets, since the target names are not file names that the targets will create. See [here](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html) for more info."_ - @grdryn circa 2023 (colorized) ([taken from here](https://github.com/opendatahub-io/ai-edge/pull/55#discussion_r1314857565))

Also,

_"This looks like a fine approach :+1: 
One possible alternative to using a ~~ConfigMap~~ `sed` would be an [OpenShift Template](https://docs.openshift.com/container-platform/4.13/openshift_images/using-templates.html) resource. [...] I think on OpenShift, a Template would be a more natural wrapper for resource definitions in the cluster that gets parameterized before being applied, but this approach is also fine for now."_ - @grdryn circa 2023 (colorized) ([taken from here](https://github.com/opendatahub-io/ai-edge/pull/55#discussion_r1314977981))

pipelines/tekton/azureml-container-pipeline/model-upload/basic-pod.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: basic-pod
+spec:
+  volumes:
+    - name: PVC_NAME
```
piotrpdev, line 7: _"I know this is just a placeholder, but it's not a valid name for an object in Kubernetes, because of the `_`s. See here for more info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names"_ - @grdryn circa 2023 (colorized) ([taken from here](https://github.com/opendatahub-io/ai-edge/pull/55#discussion_r1315092519)) (not sure if this one still applies here)

pipelines/tekton/azureml-container-pipeline/model-upload/basic-pod.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: basic-pod
```
piotrpdev, line 4: I feel like this name is a bit too generic, maybe `local-model-to-pv-pod` or something like that would be better.

pipelines/tekton/azureml-container-pipeline/model-upload/basic-pod.yaml
```diff
@@ -0,0 +1,17 @@
...
+spec:
+  volumes:
+    - name: PVC_NAME
+      persistentVolumeClaim:
+        claimName: PVC_NAME
+  containers:
+  - name: basic-container
```
piotrpdev, line 11: This is also probably too generic.

piotrpdev (commented):
Good job so far 👍

piotrpdev:
Also in the A/C in the issue (#87) I specified we should provide a template PVC so please do that.

pipelines/tekton/azureml-container-pipeline/model-upload/basic-pod.yaml
```diff
@@ -0,0 +1,17 @@
...
+    image: ubi9
```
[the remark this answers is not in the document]
jackdelahunt: 👍

pipelines/tekton/azureml-container-pipeline/model-upload/basic-pod.yaml
jackdelahunt: 👍

pipelines/tekton/azureml-container-pipeline/model-upload/basic-pod.yaml
jackdelahunt: 👍

pipelines/tekton/azureml-container-pipeline/model-upload/basic-pod.yaml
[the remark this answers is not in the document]
jackdelahunt: I tried this but with the micro version `tar` isn't installed so using `oc cp` doesn't work

pipelines/tekton/azureml-container-pipeline/model-upload/basic-pod.yaml
jackdelahunt: Done 👍  changed to just `PVC`

jackdelahunt:
> Also in the A/C in the issue (#87) I specified we should provide a template PVC so please do that eventually.

How will this work, will it be a template used for the pipelines aswell or just come with this work that you can then add in the pipelineRun?

pipelines/tekton/azureml-container-pipeline/model-upload/basic-pod.yaml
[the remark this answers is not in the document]
piotrpdev: Just checked and it doesn't come with ubi-minimal either 😕 fair enough then.

piotrpdev:
> Is the idea to create the PVC and the pod at the same time and upload the model and use this PVC to them use or copy it to somewhere else? My thought was that we just pass a PVC name that is already in the project.

I added some more detail to the issue yesterday, my bad for it being unclear, https://github.com/opendatahub-io/ai-edge/issues/87#issuecomment-1745780125

piotrpdev:
This is also a diagram for what I'm thinking:

image

LaVLaS (commented):
Can you rename the templates file so they are clearly djust to show that they are templates.

local-model-pvc-template.yaml
local-model-to-pvc-pod-template.yaml

pipelines/tekton/azureml-container-pipeline/model-upload/local-model-to-pvc-pod.yaml
```diff
@@ -0,0 +1,29 @@
...
+    containers:
+    - name: local-model-to-pvc-container
+      image: ubi9
+      stdin: true
+      tty: true
+      securityContext:
+        allowPrivilegeEscalation: false
+      volumeMounts:
+          - mountPath: /workspace/${PVC}
+            name: ${PVC}
```
grdryn, line 24: I think you could consider putting this into the object list of the other template, if they're always expected to be processed together, and the `PVC` parameter will always have the same value. WDYT?

pipelines/tekton/azureml-container-pipeline/model-upload/local-model-pvc.yaml
```diff
@@ -0,0 +1,24 @@
...
+        storage: ${SIZE} #1Gi
+parameters:
+- name: PVC
+  description: Name of PVC to be mounted to pod
+  from: '[A-Z0-9]{8}'
+  generate: expression
+- name: SIZE
+  description: Size of the PVC
+  from: '[A-Z0-9]{8}'
+  generate: expression
```
grdryn, line 24: The combination of `generate: expression` and `from` fields in each parameter definition here will cause a random value to be generated matching the `from` value, if no parameter is passed explicitly. I don't think this is appropriate for the `SIZE` parameter, and may not be for the `PVC` one either, but I'm less sure about that.

The following might be a better way to have more sane default/fallback values (which double as hints for the format of expected values, which I guess is the intention of the comment on L15):

```suggestion
- name: PVC
  description: Name of PVC to be mounted to pod
  value: local-model
- name: SIZE
  description: Size of the PVC
  value: 1Gi
```

jackdelahunt: Thanks I made that change 👍

pipelines/tekton/azureml-container-pipeline/model-upload/local-model-to-pvc-pod.yaml
jackdelahunt: The PVC and the pod want to be deleted at different times, this is because when you upload the model file to the PVC it is mounted to the pod. For another pod to mount the PVC the pod we create must un-mount it as you can't have both mounting at the same time.

So when you are deleting them you want to do it separately.

pipelines/tekton/azureml-container-pipeline/model-upload/local-model-to-pvc-pod-template.yaml
```diff
@@ -0,0 +1,29 @@
...
+      securityContext:
+        allowPrivilegeEscalation: false
+      volumeMounts:
+          - mountPath: /workspace/${PVC}
+            name: ${PVC}
+parameters:
+- name: PVC
+  description: Name of PVC to be mounted to pod
+  from: '[A-Z0-9]{8}'
+  generate: expression
```
grdryn, line 29:
```suggestion
  value: local-model
```

^to match the other one

pipelines/tekton/azureml-container-pipeline/model-upload/Makefile
```diff
@@ -0,0 +1,23 @@
...
+.PHONY: upload_model
+upload_model:
+	@oc exec local-model-to-pvc-pod -- mkdir workspace/${PVC}/model_dir/pipelines/models/ -p
+	@oc cp ${MODEL_PATH} local-model-to-pvc-pod:/workspace/${PVC}/model_dir/pipelines/models/${NAME}
+
+# deletes the pod created from the template
+# PVC - name of he PVC that is mounted to the pod
+.PHONY: clean
+clean:
+	@oc process -f local-model-to-pvc-pod.yaml -p PVC=${PVC} | oc delete -f -
```
grdryn, line 23: Given that the pod name is not currently configurable in the template, this might be slightly simpler (and at that point, a dedicated `make` target might be unnecessary?):

```suggestion
	@oc delete pod local-model-to-pvc-pod
```

pipelines/tekton/build-container-image-pipeline/check-model-and-containerfile-exists.yaml
```diff
@@ -12,14 +12,16 @@ spec:
...
+    type: string
   steps:
   - name: check-model-and-containerfile
     image: registry.access.redhat.com/ubi9/ubi-micro
     script: |
       #!/usr/bin/env bash
 
       # Check model files exist
-      ls -l $(params.model-name)/
+      ls -l model_dir/$(params.modelRelativePath)/$(params.model-name)/
```
grdryn, line 24: Something to keep in mind for future: we should decide on which casing we want to use for params here, and use that consistently (i.e. paramName vs param-name vs other varieties).

adelton:
Please note that the use of `model_dir` seems to cause https://github.com/opendatahub-io/ai-edge/issues/175.

pipelines/tekton/azureml-container-pipeline/model-upload/Makefile
adelton, line 1: To avoid expanding the problem of https://github.com/opendatahub-io/ai-edge/issues/110, if the Makefile is indeed needed, can you please describe its use in the documentation?

pipelines/tekton/azureml-container-pipeline/model-upload/local-model-pvc-template.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: template.openshift.io/v1
+kind: Template
```
adelton, line 2: Given the rest of the repository uses Kustomize, is introduction of this OpenShift specific `Template` really needed?

pipelines/tekton/azureml-container-pipeline/model-upload/local-model-pvc-template.yaml
adelton, line 1: Please note that we renamed the directories in https://github.com/opendatahub-io/ai-edge/pull/131 so you might want to rebase on `main`.

adelton (changes requested):

pipelines/tekton/azureml-container-pipeline/model-upload/local-model-to-pvc-pod.yaml
adelton: Do we really want to rely on the same PVC being remounted for various uses?

Shouldn't the Pod writing the model to the PV use one PVC, and then Pods consuming the model different PVC(s), possibly with different selectors for labels set by the populating Pod after it has finished storing the model?

pipelines/tekton/azureml-container-pipeline/model-upload/Makefile
jackdelahunt: 👍

pipelines/tekton/azureml-container-pipeline/model-upload/Makefile
jackdelahunt: Sure I will do this, but on another note should the name of the pod be configurable?

pipelines/tekton/azureml-container-pipeline/model-upload/local-model-to-pvc-pod-template.yaml
jackdelahunt: 👍

pipelines/tekton/build-container-image-pipeline/check-model-and-containerfile-exists.yaml
jackdelahunt: Yeah I agree, I mentioned this is Srijoni's PR. I didn't want to go changing other things though 😄

pipelines/tekton/azureml-container-pipeline/model-upload/local-model-to-pvc-pod.yaml
jackdelahunt: Yes the pod writing the model does only use one PVC, it is mounted then `oc cp` is used and then unmounted and the pod is deleted.

For our usecase the PVC we upload to is then used as a workspace in the pipelines, so the PVC it uses for the pipelines can copy over the model file. So I think adding labels etc. isn't needed for this. (Atleast I don't think so)

pipelines/tekton/azureml-container-pipeline/model-upload/local-model-pvc-template.yaml
jackdelahunt: Done 👍

pipelines/tekton/azureml-container-pipeline/model-upload/Makefile
jackdelahunt: > the Pod isn't actually important, and can be deleted once the model has been uploaded, is that right?

Yes the pod is just there to use `oc cp` and is not important

jackdelahunt: > maybe a generated name (or a name based on the PVC name, like ${PVC}-upload) 

Yes that would be good I think, but right now because they are in different targets that might be a pain but if we figure out that I would be happy to make that change

jackdelahunt: > recipe, rather than having them spread out in different places? 

This was what I wanted first, do it all in one make target. I couldn't figure out how to wait for the pod to be fully created before going on to the next command though (I didn't go too deep, so it is probably possible). 

If we can get something like `create pod, create PVC -> mount PVC -> oc cp -> delete pod` all in one make target that would be great

pipelines/tekton/azureml-container-pipeline/model-upload/local-model-pvc-template.yaml
jackdelahunt: Yeah maybe using Kustomize is better, when assigned this it was always said to use `Template` but now that you say it I am not sure why

adelton: If you have run-time variables / parameters in mind, then no: https://kubectl.docs.kubernetes.io/faq/kustomize/eschewedfeatures/#build-time-side-effects-from-cli-args-or-env-variables. The run should be reproducible with the content in the files, not something specified on a command line or in environment variables.

But you can of course override things with kustomize (including using patches or replacements), and that override of default can be as minimal as useful.

pipelines/tekton/azureml-container-pipeline/model-upload/Makefile
grdryn: There's a `kubectl wait` subcommand, so I think what you're looking for is something like this, is it? I haven't tried it out:

```bash
@oc process -f local-model-to-pvc-pod-template.yaml -p PVC=${PVC} | oc create -f -
@oc wait pod local-model-to-pvc-pod --for condition=Ready=True --timeout=120s
@oc exec local-model-to-pvc-pod -- mkdir -p workspace/${PVC}/model_dir/pipelines/models/
@oc cp ${MODEL_PATH} local-model-to-pvc-pod:/workspace/${PVC}/model_dir/pipelines/models/${NAME}
@oc delete pod local-model-to-pvc-pod
```

jackdelahunt: Done 👍

grdryn: Nice! Is the clean target still needed now, or could the pod be deleted immediately after the copy in the create target?

pipelines/tekton/azureml-container-pipeline/model-upload/local-model-pvc-template.yaml
jackdelahunt: I will set defaults in the Makefile for each variable similar to how I have done here. So now each option is optional when running the make target

https://github.com/opendatahub-io/ai-edge/pull/176/files#diff-b6d53e8f4a83a8b4900586623befbee132db0973ed649fb23ea48b0754f2528aR1-R5

biswassri:
@jackdelahunt It might be a good idea to update the ReadMe file with steps for anyone trying this?  Even just the make command and some details about local pvc upload I think might help.

jackdelahunt:
@biswassri sure will do

pipelines/tekton/azureml-container-pipeline/model-upload/Makefile
grdryn: Marking this as unresolved as I still have this question above. :slightly_smiling_face:

pipelines/tekton/azureml-container-pipeline/model-upload/Makefile
```diff
@@ -0,0 +1,32 @@
+# Used for uploading local model files to the cluster through a 
+# PVC. Use create to create the pod and PVC needed, then use
+# upload model to upload a local file the PVC that the pod
+# is mounted to
+
+SIZE=1G
+PVC=model-upload-pvc
```
grdryn, line 7: When adding information to the README about the targets here, don't forget to document the variables and how to override them.

jackdelahunt: Done 👍

grdryn (approved):
LGTM, thanks @jackdelahunt!

adelton:
Hello @jackdelahunt, was there a specific ready why the **Using local models in pipelines** documentation section went to the top-level `README.md` and not to `pipelines/README.md`, seeing the whole PR was in the `pipelines/` space?

Issue #90, opened 2023-09-21
There are no provided gitops-git-user- files


adelton:
The [Git Repository and Credentials](https://github.com/opendatahub-io/ai-edge/blob/main/pipelines/README.md#git-repository-and-credentials) section says:

> Change the provided `gitops-git-user-` files to match your Git credentials

However, running
```
find . -name 'gitops-git-user-*'
```
does not find any such files.

Issue #92, opened 2023-09-21
commit-image-ref does not respect target-imagerepo


adelton:
When checking what `commit-image-ref` of [Deploy the GitOps pipeline](https://github.com/opendatahub-io/ai-edge/blob/main/pipelines/README.md#deploy-the-gitops-pipeline-1) created in `pipeline_<UUID>` branch in my Git repository, I see

```
diff --git a/acm/odh-edge/apps/tensorflow-housing-app/kustomization.yaml b/acm/odh-edge/apps/tensorflow-housing-app/kustomization.yaml
index c3ed7b7..77c7090 100644
--- a/acm/odh-edge/apps/tensorflow-housing-app/kustomization.yaml
+++ b/acm/odh-edge/apps/tensorflow-housing-app/kustomization.yaml
@@ -59,4 +59,4 @@ replacements:
 images:
 - name: edge-model-template-image
   newName: quay.io/rhoai-edge/tensorflow-housing
-  digest: sha256:b5f11c2f380e76bb4dc5b66e40fff46ee277cd9a7477a6519b95dc8017f67852
+  digest: sha256:19e7312d8cbfcbf21dcab083f9ef99af2f807f156b1216a423267010b7fdd7c0
```

Since the `skopeo-copy` task from the previous step copied the image to my Quay namespace configured by `target-imagerepo`, I would expect this `commit-image-ref` to apply that same value to `newName`. The image `sha256:19e7312d8cbfcbf21dcab083f9ef99af2f807f156b1216a423267010b7fdd7c0` does not exist in `quay.io/rhoai-edge/tensorflow-housing` if it was pushed to a different namespace.

adelton:
Hmmm, so `sha256:19e7312d8cbfcbf21dcab083f9ef99af2f807f156b1216a423267010b7fdd7c0` actually does exist in `quay.io/rhoai-edge/tensorflow-housing`. But it looks like the `gitops-update-pipeline` merely copies information from `pipelines/tekton/gitops-update-pipeline/example-pipelineruns/gitops-update-pipelinerun-tensorflow-housing.yaml` to `acm/odh-edge/apps/tensorflow-housing-app/kustomization.yaml`.

But wasn't the the goal of the exercise to actually push to Quay and record in the PR the image that was built in `azureml-container`'s `build-mlflow-container` task?

adelton:
[...]

PR #123, opened 2023-10-04
Expand description of the GitOps pipeline, update defaults to GitHub.

adelton:
## Description

Expand description of the GitOps pipeline, update defaults to GitHub.

Fixes https://github.com/opendatahub-io/ai-edge/issues/90.

## How Has This Been Tested?

I tested in my fork in GitHub until I got the pull request properly filed.

## Merge criteria:

### DIFF
--- a/pipelines/tekton/gitops-update-pipeline/example-pipelineruns/example-git-credentials-secret.yaml
+++ b/pipelines/tekton/gitops-update-pipeline/example-pipelineruns/example-git-credentials-secret.yaml
@@ -1,18 +1,18 @@
 apiVersion: v1
 kind: Secret
 metadata:
-  name: EXAMPLE_SECRET_NAME
+  name: edge-user-1
 type: Opaque
 stringData:
   # This "token" field is used for interacting with the GitHub API to
   # create a pull request. Change the value to a token from your
   # GitHub-compatible forge.
-  token: EXAMPLE_API_TOKEN
+  token: {github_pat_1234567890ABCDAPI_TOKEN}
   # This .git-credentials field is used to specify credentials when
   # interacting with a Git server (clone/fetch/push). It will be
   # placed as a file on disk so that the Git CLI can use it, so change
   # it to appropriate details for your Git server.
-  .git-credentials: http://EXAMPLE_GIT_USER:EXAMPLE_API_TOKEN@EXAMPLE_GIT_SERVER
+  .git-credentials: https://{username}:{github_pat_1234567890ABCDAPI_TOKEN}@github.com
   .gitconfig: |
     [credential]
       helper = store
--- a/pipelines/README.md
+++ b/pipelines/README.md
@@ -170,43 +170,36 @@ checksum as the one shown in the Image Stream.
 
 ## Deploy the GitOps pipeline
 
+We have so far worked with the upstream source repo.
+
+In this last pipeline we will use a clone of that repo to show how the identifier of the newly built and tested container image
+can be recorded in the repository via a pull request. For that we will need a clone or copy of this repository, in GitHub or
+in your Git server.
+
 ### Git Repository and Credentials
 
 - Clone/Mirror this repository on your Git server
-- Change the provided `gitops-git-user-` files to match your Git credentials
+- Update the `git*` parameters in
+  [`gitops-update-pipelinerun-bike-rentals.yaml`](tekton/gitops-update-pipeline/example-pipelineruns/gitops-update-pipelinerun-bike-rentals.yaml)
+  and/or
+  [`gitops-update-pipelinerun-tensorflow-housing.yaml`](tekton/gitops-update-pipeline/example-pipelineruns/gitops-update-pipelinerun-tensorflow-housing.yaml)
+  to match location of your repository and the target branch for the pull request.
+  The defaults assume `https://github.com/username/ai-edge-gitops` and `main`.
+- Update the [`example-git-credentials-secret.yaml`](tekton/gitops-update-pipeline/example-pipelineruns/example-git-credentials-secret.yaml)
+  with your repository information and credentials.
+  For GitHub, the token can be generated at Settings > Developer Settings > Personal access tokens > Fine-grained tokens
+  and it should have Read access to metadata and Read and Write access to code and pull requests permissions to the repository you use.
 
-### Deploy the GitOps pipeline
+### Deploy and run the GitOps pipeline
 
 ```bash
 oc apply -k tekton/gitops-update-pipeline/
-```
-
-#### Run the GitOps pipeline
-
-The `tekton/gitops-update-pipeline/example-pipelineruns/` contains some examples that can be modified and used.
-In these examples, notice that there is a template Secret file for the Git credentials that are referenced by different tasks.
-Create an equivalent Secret with appropriate details for your environment, and change the parameter values in the PipelineRun definition to match.
-
-NOTE: If you are using a branch other than `main` for your gitops workflow, you will need to modify the `PipelineRuns` examples under `tekton/gitops-update-pipeline/example-pipelineruns/` to set the `gitRepoBranchBase` parameter to your custom branch name
-```
-spec:
-  params:
-  - name: gitRepoBranchBase
-    value: <CUSTOM BRANCH NAME>
-```
-
-``` bash
-# Bike rentals app
-cp pipelines/tekton/gitops-update-pipeline/example-pipelineruns/example-git-credentials-secret.yaml /tmp/gitea-edge-user-1-secret.yaml
-$EDITOR /tmp/gitea-edge-user-1-secret.yaml # make changes for your specific environment
-oc apply -f /tmp/gitea-edge-user-1-secret.yaml
-
-oc create -f tekton/gitops-update-pipeline/example-pipelineruns/gitops-update-pipelinerun-bike-rentals.yaml
-
-# Tensorflow housing app
-cp pipelines/tekton/gitops-update-pipeline/example-pipelineruns/example-git-credentials-secret.yaml /tmp/gitea-edge-user-2-secret.yaml
-$EDITOR /tmp/gitea-edge-user-2-secret.yaml # make changes for your specific environment
-oc apply -f /tmp/gitea-edge-user-2-secret.yaml
-
+oc apply -f tekton/gitops-update-pipeline/example-pipelineruns/example-git-credentials-secret.yaml
 oc create -f tekton/gitops-update-pipeline/example-pipelineruns/gitops-update-pipelinerun-tensorflow-housing.yaml
+# and/or
+oc create -f tekton/gitops-update-pipeline/example-pipelineruns/gitops-update-pipelinerun-bike-rentals.yaml
 ```
+
+After the Pipeline Run(s) finish, check your git repository -- there should be a pull request with an update of the respective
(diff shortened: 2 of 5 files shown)


adelton:
While I feel this change is a valid improvement of both the YAMLs and the documentation, in https://github.com/opendatahub-io/ai-edge/issues/92#issuecomment-1747202159 I ask if the "GitOps" pipeline which only does the push + file pull request should be a separate pipeline at all, or should be a part of the `test-mlflow-image-pipeline` Pipeline.

adelton:
I've tested that with Gitea with HTTP (not HTTPS), apart from the obvious customizations like `gitServer`, `gitOrgName`, or `gitRepoName` in PipelineRun YAMLs, and the token / credentials in [`example-git-credentials-secret.yaml`](pipelines/tekton/gitops-update-pipeline/example-pipelineruns/example-git-credentials-secret.yaml),
```diff
   - name: gitApiServer
-    value: api.github.com
+    value: http://gitea-ai-edge-gitea.apps.rosa.adelt1-aisrhods.g9ax.p3.openshiftapps.com
+  - name: gitApiPrefix
+    value: /api/v1
```
is also needed. The https://github.com/tektoncd/catalog/blob/main/task/github-open-pr/0.2/github-open-pr.yaml#L127 logic "This is for our fake github server" for `http://` URLs then works correctly, hitting the `http.client.HTTPConnection` path.

The question then is, should we even support non-HTTPS deployments?

Another change needed is that `gitApiPrefix`. I wonder if we should have it `default: ''`` in [gitops-update-pipeline.yaml](pipelines/tekton/gitops-update-pipeline/gitops-update-pipeline.yaml) or if we should require its explicit use in the PipelineRun YAML.

pipelines/tekton/gitops-update-pipeline/example-pipelineruns/example-git-credentials-secret.yaml
piotrpdev, line 1: @grdryn I don't know if now that I've discovered the [OpenShift Templates](https://docs.openshift.com/container-platform/4.13/openshift_images/using-templates.html) hammer everything is a nail, but how about using one for this file? Probably better for security too in case the user accidentally commits this file with their credentials in them.

piotrpdev (commented):
The default should probably be the more common option ¯\\_(ツ)_/¯, @LaVLaS which do you think is more popular for our use case: on-prem Git server or GitHub?

pipelines/tekton/gitops-update-pipeline/example-pipelineruns/example-git-credentials-secret.yaml
LaVLaS: In real world scenarios, on-prem git server (from any provider) would be used for security purposes.  

For this PoC, I would default to `github.com/opendatahub-io/ai-edge` but only as a marker text for `sed` to replace it with a custom url.  We can simplify the setup and configuration if we make an in-cluster gitea server a dependency to minimize the risk and configuration when using a public repo on github.


> I don't know if now that I've discovered the OpenShift Templates hammer everything is a nail, but how about using one for this file?
I forget that OpenShift templates are still around since there have been discussions to deprecate it since the early releases of OCP <= v4.5.  It was never accepted upstream (same as DeploymentConfig) and they have been superseded by workflow improvements (kustomize, helm, k8s native application development, ...)

adelton: If we talk about security, using a private GitHub repo over HTTPS seems a bit more secure than an automatically-deployed Gitea with HTTP and well-known admin password.

grdryn: > > I don't know if now that I've discovered the OpenShift Templates hammer everything is a nail, but how about using one for this file?
>
> I forget that OpenShift templates are still around since there have been discussions to deprecate it since the early releases of OCP <= v4.5.  It was never accepted upstream (same as DeploymentConfig) and they have been superseded by workflow improvements (kustomize, helm, k8s native application development, ...)

@LaVLaS For use cases where we want to have users provide specific values at run time to make a generic template specific to their case, do you think that we should avoid OpenShift Templates in favour of other things like Helm templates (I haven't used these, but they seem somewhat analogous), or just provide instructions on how to make the resources they need by replacing values manually or using something like sed?

OTOH there's nothing in the [OpenShift docs](https://docs.openshift.com/container-platform/4.13/openshift_images/using-templates.html#templates-overview_using-templates) about templates being deprecated yet, and even if they were, presumably there would be a new major release before they'd be actually removed?

----

@LaVLaS says:
> In real world scenarios, on-prem git server (from any provider) would be used for security purposes.

then @adelton says:
> If we talk about security, using a private GitHub repo over HTTPS seems a bit more secure than an automatically-deployed Gitea with HTTP and well-known admin password.

I wonder how difficult it would be to deploy Gitea with HTTPS and unique passwords? We could also configure the repos on it to be private. Would that be a worthwhile improvement? It might be a better demo, and have slightly fewer manual setup steps than needing to get GitHub tokens, etc.

adelton: Don't get me wrong, I like that Gitea exists. And we might want to test with it and show how to set it up. But then the question might be why Gitea and not GitLab or Gitbucket or something else.

If you come to try the PoC to see what the setup is about, you might want to see as few mandatory pieces as possible, to get the core message of the PoC. And given GitHub's prominence, people who might come to the repo are pretty likely to already have a GitHub account.

LaVLaS: > do you think that we should avoid OpenShift Templates in favour of other things like Helm templates

Helm chart is always the better option as it has become the industry standard across all kubernetes distributions for templating manifests.  I don't know if we want to invest heavily into an ai-edge helm chart as the original goal was to investigate RHOAI in Edge scenarios  and the there are other projects internally and externally that can provision an AI Edge environment from scratch

> ...or just provide instructions on how to make the resources they need by replacing values manually or using something like sed?

My vote is for this simple lightweight workflow as we only need to provide guidance for adding a few zludx that are easily supported by walking the directory tree with `sed` to replace values.  

> OTOH there's nothing in the OpenShift docs about templates being deprecated yet, and even if they were, presumably there would be a new major release before they'd be actually removed?

It's been so long since I used OpenShift templates that the only thing I remember is that they were not adopted by the upstream community.  It looks like they are still in use for OpenShift sample applications

```
oc get templates -n openshift
```

adelton:
Rebased on main.

adelton:
Rebased on main.

adelton:
https://github.com/opendatahub-io/ai-edge/pull/137 builds on top of this, so I'd need a decision about this pull request to proceed with review of that one.

Unless we are able to provide steps / YAMLs to create secure Gitea setups pretty quickly, I'd urge to go with this GitHub approach, at least for now.

I've been testing this and https://github.com/opendatahub-io/ai-edge/pull/137 with a private GitHub repo and with a fine-grained personal access token for multiple days now, I'm fairly confident about both its security and ease of use.

grdryn (approved):
Sorry for the delay, this was on my TODO list to get back to. LGTM, thanks!

pipelines/tekton/gitops-update-pipeline/example-pipelineruns/example-git-credentials-secret.yaml
```diff
@@ -1,18 +1,18 @@
...
 metadata:
-  name: EXAMPLE_SECRET_NAME
+  name: edge-user-1
 type: Opaque
 stringData:
   # This "token" field is used for interacting with the GitHub API to
   # create a pull request. Change the value to a token from your
   # GitHub-compatible forge.
-  token: EXAMPLE_API_TOKEN
+  token: github_pat_1234567890ABCDAPI_TOKEN
```
LaVLaS, line 10: Since the previous config values (username, gitServer, ...) where using actual values that worked in our demo cluster, should we update the new values that we expect the user to change to be something that will break the manifest deployment if not updated for the custom environment?

adelton: I'm not sure I follow.

The `EXAMPLE_API_TOKEN` hopefully was not a valid token, nor is `github_pat_1234567890ABCDAPI_TOKEN`. At the same time, with that `github_pat_*` format I try to show the user what they should be after in their GitHub UI.

adelton:
Fixes https://github.com/opendatahub-io/ai-edge/issues/90.

pipelines/tekton/gitops-update-pipeline/example-pipelineruns/example-git-credentials-secret.yaml
LaVLaS: Just to clarify.

The references to a replacement variable like `github_pat_1234567890ABCDAPI_TOKEN` would be changed to  `<GITHUB_PAT_1234567890ABCDAPI_TOKEN>`.  This would make it obvious to the the user that it needs to be updated AND if unchanged it would clearly break in a way to makes it obvious that this value was still applied to the cluster

adelton: I still don't understand what you suggest. Adding the `<` and `>` there seems to be valid YAML, so `oc apply` still passes:
```bash
$ git diff
diff --git a/pipelines/tekton/gitops-update-pipeline/example-pipelineruns/example-git-credentials-secret.yaml b/pipelines/tekton/gitops-update-pipeline/example-pipelineruns/example-git-credentials-secret.yaml
index b12f7b5..d14b8dc 100644
--- a/pipelines/tekton/gitops-update-pipeline/example-pipelineruns/example-git-credentials-secret.yaml
+++ b/pipelines/tekton/gitops-update-pipeline/example-pipelineruns/example-git-credentials-secret.yaml
@@ -7,7 +7,7 @@ stringData:
   # This "token" field is used for interacting with the GitHub API to
   # create a pull request. Change the value to a token from your
   # GitHub-compatible forge.
-  token: github_pat_1234567890ABCDAPI_TOKEN
+  token: <github_pat_1234567890ABCDAPI_TOKEN>
   # This .git-credentials field is used to specify credentials when
   # interacting with a Git server (clone/fetch/push). It will be
   # placed as a file on disk so that the Git CLI can use it, so change
$ oc apply -f pipelines/tekton/gitops-update-pipeline/example-pipelineruns/example-git-credentials-secret.yaml
secret/edge-user-1 created
```

I mean, I can add those there but it does not seem it would achieve the clear break that you are after. :-)

LaVLaS: Apologies. I had to remember what characters would  break an unquoted yaml string.  It should be `token: {{github_pat_1234567890ABCDAPI_TOKEN}} .  `kustomize` should throw an error

```
Error: map[string]interface {}(nil): yaml: invalid map key: map[string]interface {}{"github_pat_1234567890ABCDAPI_TOKEN":""}
```

adelton: That works. PR updated to 1a573158c97a7809a27b7be9a6a47586fb43b771.

piotrpdev (approved):

adelton:
Thanks, merging.

Issue #125, opened 2023-10-04
Make GitOps pipeline use the latest tested image and repo


piotrpdev:
## Description

The current pipeline workflow requires the user to manually copy the resulting image digest of a newly tested and uploaded image from the testing pipeline:

https://github.com/opendatahub-io/ai-edge/blob/f148f63a4c4288ad917428e63f6bfeac2f5355a2/pipelines/tekton/test-mlflow-image-pipeline/test-mlflow-image-pipeline.yaml#L120-L121

to the GitOps pipeline.

https://github.com/opendatahub-io/ai-edge/blob/f148f63a4c4288ad917428e63f6bfeac2f5355a2/pipelines/tekton/gitops-update-pipeline/gitops-update-pipeline.yaml#L13-L14

The GitOps pipeline also doesn't validate the image repo, the pipeline should be modified to:

- By default, having a step in the GitOps pipeline that uses the [`openshift-client` ClusterTask](https://hub.tekton.dev/tekton/task/openshift-client) to get the latest successfully tested and uploaded image ([like so](https://github.com/opendatahub-io/ai-edge/issues/68#issuecomment-1747485029)) along with the repo to which it was uploaded, and make the PR with that.

- Optionally, allow the user to manually specify the image repo and digest so they can override the values from above.

> **Warning**
> Make sure you're getting the latest image for the **correct model**, this will require adding a `"model-name"` and `"model-version"` param to the GitOps pipeline similar to [the other pipelines](https://github.com/opendatahub-io/ai-edge/blob/f148f63a4c4288ad917428e63f6bfeac2f5355a2/pipelines/tekton/azureml-container-pipeline/azureml-container-pipeline.yaml#L7-L11).

> **Note**
> Modifying the testing pipeline to [emit the uploaded image repo and digest as a result](https://tekton.dev/docs/pipelines/pipelines/#emitting-results-from-a-pipeline) is probably a better approach than using a long and complicated `jsonpath` in the `oc` command.

Resolves #92 

## A/C
- Resulting testing pipeline data doesn't have to be copied to the GitOps pipeline params for it to work
- GitOps pipeline:
	- Uses image repo and digest from the latest successfully tested and uploaded image **(for the correct model)** by default
	- Allows the user to override the values above using the params if they want to
- All of the pipelines and the standard workflow in [pipeline/README](https://github.com/opendatahub-io/ai-edge/blob/main/pipelines/README.md) still works (can't be too safe)

[...]

PR #135, opened 2023-10-06
Add gitops-update Pipeline parameter to support a custom target git branch

LaVLaS:
## Description
Introduce a parameter to the `gitop-update-pipeline` to allow users to specify a target branch name for the automated PR pipeline.  The default value is still `main` but this will allow a `PipelineRun` to target custom branch like `my-edge-demo`

## How Has This Been Tested?
I deployed this fix locally as part of the ai-edge demo that was using a new branch that was created from `main`

## Merge criteria:
No issues with the parameter and the documentation is clear

- [ ] The commits are squashed in a cohesive manner and have meaningful messages.
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has manually tested the changes and verified that the changes work

PR #137, opened 2023-10-12
File the pull request with information about the latest built and pushed image for the model

adelton:
## Description

File the pull request with information about the latest built and pushed image for the model retrieved from the PipelineRuns.

Fixes https://github.com/opendatahub-io/ai-edge/issues/92.

Fixes https://github.com/opendatahub-io/ai-edge/issues/125.

This builds on top of https://github.com/opendatahub-io/ai-edge/pull/123.

## How Has This Been Tested?

Manually.

## Merge criteria:

### DIFF
--- a/pipelines/tekton/gitops-update-pipeline/test-mlflow-retrieve-image-info-task.yaml
+++ b/pipelines/tekton/gitops-update-pipeline/test-mlflow-retrieve-image-info-task.yaml
@@ -1,0 +1,20 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: retrieve-image-info
+spec:
+  description: Retrieve SHA of the built image from the previous PipelineRun
+  params:
+  - name: namespace
+    type: string
+  - name: model-name
+    type: string
+  steps:
+  - name: get-image-sha
+    image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
+    script: |
+      oc get -n $(params.namespace) pipelinerun --selector tekton.dev/pipeline=test-mlflow-image --sort-by=.status.completionTime -o jsonpath='{range .items[?(@.spec.params[0].name == "model-name")]}{.spec.params[?(@.name == "model-name")].value} {.status.results[?(@.name == "image-sha")].value}{"\n"}{end}' | awk -v model=$(params.model-name) '$1 == model && NF == 2 { print $2 }' | tail -1 | tee /dev/stderr | while read sha ; do echo -n "$sha" > $(results.image-sha.path) ; done ;
+      test -s $(results.image-sha.path)
+  results:
+  - name: image-sha
+    description: The image checksum
--- a/pipelines/tekton/gitops-update-pipeline/test-mlflow-retrieve-image-info-task.yaml
+++ b/pipelines/tekton/gitops-update-pipeline/test-mlflow-retrieve-image-info-task.yaml
@@ -3,7 +3,7 @@ kind: Task
 metadata:
   name: retrieve-image-info
 spec:
-  description: Retrieve SHA of the built image from the previous PipelineRun
+  description: Retrieve location and SHA of the built image from the previous PipelineRun
   params:
   - name: namespace
     type: string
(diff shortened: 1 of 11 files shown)


pipelines/tekton/gitops-update-pipeline/test-mlflow-retrieve-image-info-task.yaml
```diff
@@ -0,0 +1,22 @@
...
+  params:
+  - name: namespace
+    type: string
+  - name: model-name
+    type: string
+  steps:
+  - name: get-image-sha
+    image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
+    script: |
+      oc get -n $(params.namespace) pipelinerun --selector tekton.dev/pipeline=test-mlflow-image --sort-by=.status.completionTime -o jsonpath='{range .items[?(@.spec.params[0].name == "model-name")]}{.spec.params[?(@.name == "model-name")].value} {.status.results[?(@.name == "image-sha")].value} {.status.results[?(@.name == "target-registry-url")].value}{"\n"}{end}' | awk -v model=$(params.model-name) '$1 == model && NF == 3 { print $2, $3 }' | tail -1 | tee /dev/stderr | while read sha registry ; do echo -n "$sha" > $(results.image-sha.path) ; echo -n "$registry" > $(results.target-registry-url.path) ; done ;
```
LaVLaS, line 16: Since a Task is supposed to be a reusable self-contained action, I think this script should stay embedded in the `Pipeline` since it is hardcoded to `test-mlflow-image`.

Instead of querying a `PipelineRun`, can we simplify this workflow to rely on a known incluster ImageTag OR a ConfigMap that holds the current `ImagestreamTag` to reference for this Pipeline step

adelton: The problem is, with a task defined merely as an entry in `tasks` in `Pipeline`, we are not able to have `results` there, IIRC. The only place I found `results` definition working is when I created it as a separate `Task`.

Yes, the task hardcodes `test-mlflow-image` -- but we have to hardcode something, to be able to reliably pick up the values from the previous PipelineRuns. I can turn it into a default of a parameter, if that's what you are suggesting.

The problem of using an ImageTag -- we want to be able to record in the pull request the SHA-256 of the exact image that was tested by the `test-mlflow-image` Pipeline. And the tag can change between the testing PipelineRun running the application and this PipelineRun finding the tag. That's why the whole logic of providing that value through `results` in the PipelineRuns.

Or do you suggest for the `test-mlflow-image` to set some very-well-known tag like `:testing-passed` any time the tests pass?

adelton:
Yes, we will need a rebase now that https://github.com/opendatahub-io/ai-edge/pull/135 got merged.

adelton:
Rebased on main on top of merged https://github.com/opendatahub-io/ai-edge/pull/135.

pipelines/tekton/gitops-update-pipeline/test-mlflow-retrieve-image-info-task.yaml
adelton: Of course, my ideal approach would be to do it all in one Pipeline, test and file the pull request. But @piotrpdev was not fond of that.

adelton:
Rebased on top of main.

adelton:
@piotrpdev Thanks for filing https://github.com/opendatahub-io/ai-edge/issues/146.

Now the question is -- should we rework this PR into the one-pipeline approach, or can we merge this to fix the specific problem (the recording `newName` and `digest` of the pushed container image), and treat the merging of the pipelines as a followup RFE? I likely won't be able to focus on that rework into a single pipeline in the next eigth days.

My concern at this point is to have a working end-to-end workflow, with solid documentation, as a basis for further development. So while I would happily work on https://github.com/opendatahub-io/ai-edge/issues/146 as that approach was my preference for end solution, I might propose to just merge what we have here, and build on top of that.

piotrpdev:
> So while I would happily work on #146 as that approach was my preference for end solution, I might propose to just merge what we have here, and build on top of that.

I agree with this approach 👍 .

piotrpdev (approved):

piotrpdev:
Might want to wait for other opinions to come in first before merging e.g. @LaVLaS

adelton:
Thanks, merging.

pipelines/tekton/gitops-update-pipeline/test-mlflow-retrieve-image-info-task.yaml
LaVLaS: > Or do you suggest for the test-mlflow-image to set some very-well-known tag like :testing-passed any time the tests pass?

Your intent is sound but in the event we have to pass some config between PipelineRuns then we should use a configmap or annotation on the target object.

adelton: I merged the changes after approval by @piotrpdev and after having https://github.com/opendatahub-io/ai-edge/issues/146 for the followup work. Let's consider this an interim quick-fix for the https://github.com/opendatahub-io/ai-edge/issues/125 problem and a baseline for further work.

PR #139, opened 2023-10-12
Order components based on their usage, losen version requirements.

adelton:
## Description

Order components based on their usage, losen version requirements, add information about accounts on hosted services.

## How Has This Been Tested?

Note tested.

## Merge criteria:

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,14 +2,15 @@
 
 Artifacts in support of ODH Edge use cases that show integration with Red Hat Advanced Cluster Management (Open Cluster Management)
 
-| Components                           | Version |
-|--------------------------------------|---------|
-| OpenShift                            | 4.13    |
-| Open Data Hub                        | 2.x     |
-| Red Hat Advanced Cluster Management  | 2.8     |
-| OpenShift Pipelines                  | 1.11.x  |
-| Quay Registry                        | 2.8     |
-| Gitea                                | 1.20.2  |
+| Components                                          | Version        |
+|-----------------------------------------------------|----------------|
+| OpenShift clusters (at least two) with admin access | 4.12 or higher |
+| GitHub account or Gitea installation                | [github.com](https://github.com/) |
+| AWS account with access to S3                       | [s3.console.aws.amazon.com](https://s3.console.aws.amazon.com/) |
+| Red Hat OpenShift Pipelines                         | 1.11 or higher |
+| Quay Registry account                               | [quay.io](https://quay.io/) |
+| Advanced Cluster Management for Kubernetes          | 2.8            |
+| Open Data Hub (optional)                            | 1.x or 2.x     |
 
 ## Proof of Concept Edge use case with ACM
 


README.md
```diff
@@ -2,14 +2,15 @@
...
-| Gitea                                | 1.20.2  |
+| Components                                          | Version        |
+|-----------------------------------------------------|----------------|
+| OpenShift clusters (at least two) with admin access | 4.12 or higher |
+| GitHub account or Gitea installation                | [github.com](https://github.com/) |
+| AWS account with access to S3                       | [s3.console.aws.amazon.com](https://s3.console.aws.amazon.com/) |
+| Red Hat OpenShift Pipelines                         | 1.11 or higher |
+| Quay Registry account                               | [quay.io](https://quay.io/) |
+| Advanced Cluster Management for Kubernetes          | 2.8            |
+| Open Data Hub (optional)                            | 1.x or 2.x     |
```
LaVLaS, line 13: We should drop 1.x as it is EOL

LaVLaS:
Should we move the components that are not deployed directly (Git server, Object Storage, Image Registry, ...) as part of the Edge use case but have expected configuration steps to a `Requirements` section?  There is some setup that is expected to run this use case, primarily auth credentials, that may need explanation

README.md
```diff
@@ -2,14 +2,15 @@
...
-| Open Data Hub                        | 2.x     |
-| Red Hat Advanced Cluster Management  | 2.8     |
-| OpenShift Pipelines                  | 1.11.x  |
-| Quay Registry                        | 2.8     |
-| Gitea                                | 1.20.2  |
+| Components                                          | Version        |
+|-----------------------------------------------------|----------------|
+| OpenShift clusters (at least two) with admin access | 4.12 or higher |
+| GitHub account or Gitea installation                | [github.com](https://github.com/) |
+| AWS account with access to S3                       | [s3.console.aws.amazon.com](https://s3.console.aws.amazon.com/) |
```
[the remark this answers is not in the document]
adelton: Currently the code only works with S3, doesn't it?

README.md
adelton: https://opendatahub.io/docs/quick-installation/ has not such information, and it is linked from the top-right "Get started" link at https://opendatahub.io/. So that it the only "supported" one, since 2.x is deemed Alpha.

Where is the EOL of 1.x tracked?

adelton:
For stuff needed for the pipelines part, we've been amending the https://github.com/opendatahub-io/ai-edge/blob/main/pipelines/README.md to describe what needs to be configured right there. I'd rather keep the information at the place where it is consumed because there you can also pretty clearly see if it was correct, that's why I've been adding those "check what happened" steps there.

I don't like having the setup steps all over the place, not clear what is needed for what.

adelton:
The rest of the README is being updated via https://github.com/opendatahub-io/ai-edge/pull/140.

README.md
LaVLaS: Then the information on opendatahub.io still needs to be updated but there is no need to add ODH 1.x when it was set to 2.x previously

README.md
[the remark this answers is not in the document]
LaVLaS: It only works with the 'S3` API but that api is supported by other Object Storage providers, correct?

piotrpdev (approved):

adelton:
Thanks, merging.

README.md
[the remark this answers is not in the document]
adelton: It might ... but did we test it?

README.md
adelton: Let's do a separate change removing that `1.x` once a PR in https://github.com/opendatahub-io/opendatahub.io clarifying the status of those version is in.

That should serve as a nagging tool to keep the overall information consistent.

adelton: After all, ODH is not used in the PoC at all, AFAICS. We just wave-handily say "use ODH if you want to train your model" but that really it.

I was actually thinking about removing the mention of it completely but then kept it, just made it more vague.

PR #143, opened 2023-10-12
Use digests of newest images in quay.io/rhoai-edge/ repositories.

adelton:
## Description

Use digests of newest images in https://quay.io/organization/rhoai-edge repositories.

## How Has This Been Tested?

Not tested, just copied the SHA-256's from quay WebUI.

## Merge criteria:

### DIFF
--- a/acm/odh-edge/apps/bike-rental-app/kustomization.yaml
+++ b/acm/odh-edge/apps/bike-rental-app/kustomization.yaml
@@ -76,4 +76,4 @@ replacements:
 images:
 - name: edge-model-template-image
   newName: quay.io/rhoai-edge/bike-rentals-auto-ml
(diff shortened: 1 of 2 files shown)


piotrpdev:
Wouldn't using the older `1-2` versions of the images be better so we can showcase updating the digests to `1-3`? This would also be useful for testing/showcasing the PR feature without uploading a new image.

adelton:
That makes sense. So the goal would be not to go with the newest but with the oldest?

On the other hand, you can still show how a different image gets used but going from the latest to `1-3` and to `1-2` ... so we don't lose much by going with the latest, plus things are easier to check for consistency.

adelton:
Rebased on main.

piotrpdev (approved):
> going from the latest to 1-3 and to 1-2

That's actually a more realistic scenario, fair enough 😅

adelton:
Thanks, merging.

PR #153, opened 2023-10-24
Add info for AI Edge w/ ODH

LaVLaS:
## Description
Updates the README to show the ODH version and expectations for feature development. Removes references to RHODS as the features implemented will not be available in the downstream product until they are incubated in ODH until

## How Has This Been Tested?
N/A

## Merge criteria:
Information is clear and provides a quick high level summary of ODH integration

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Open Data Hub / AI Edge Use Cases
 
-This repository contains artifacts to show and support Open Data Hub and Red Hat OpenShift Data Science Edge use cases.
+This repository contains artifacts to show and support Open Data Hub use cases.
 
 For the purpose of a "near edge" Proof of Concept, the edge environment is represented by separate OpenShift cluster(s),
 managed from a core OpenShift cluster using Red Hat Advanced Cluster Management (ACM),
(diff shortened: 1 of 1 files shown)


adelton (commented):
Given that the PoC currently works both on ODH and RHODS, is it really important to remove that RHODS reference? I mean, it is not a secret that RHODS exists, and some stakeholders and people looking at this setup might actually be happier to see that this could be set up with downstream bits, not just with ODH.

Also, you remove the note that ODH is optional ... but at this point, there is nothing ODH would be needed for, except for training the model. And that we note in https://github.com/opendatahub-io/ai-edge#developing-and-training-a-model is out of scope, and thus optional.

I don't mind removing the ODH 1.X reference but it's weird to do so when 2.X is still designated alpha in https://opendatahub.io/docs/quick-installation-new-operator/.

If I had a vote, this would be a nack.

LaVLaS:
> Given that the PoC currently works both on ODH and RHODS, is it really important to remove that RHODS reference?

The intent was not to hide RHODS. As we take the knowledge gained from the Proof of Concept and use that to develop the feature in ODH, none of the RHODS releases planned for this year will have any interations of the work that is being developed.

> but at this point, there is nothing ODH would be needed for, except for training the model.

In the context of the PoC it was optional but we are working to integrate these external workflows within ODH so the trained model can be packaged in a container, tested and pushed from within ODH

> I don't mind removing the ODH 1.X reference but it's weird to do so when 2.X is still designated alpha in

ODH 1.x is planned for deprecation with the 1.11 release (See [opendatahub-operator #648](https://github.com/opendatahub-io/opendatahub-operator/issues/648)).  The `alpha` reference in opendatahub.io is just an oversight as it was used as a preview for the downstream release that started in July.  We accelerated the development of v2.x and the opendatahub.io documentation for the ODH install just needs a refresh

> If I had a vote, this would be a nack.

Noted.  Your input is still appreciated.

adelton:
> 
> The intent was not to hide RHODS. As we take the knowledge gained from the Proof of Concept and use that to develop the feature in ODH, none of the RHODS releases planned for this year will have any interations of the work that is being developed.

Which in turn means that if you want to see the approach on RHODS, for the time being this is the latest greatest you can get.

> In the context of the PoC it was optional but we are working to integrate these external workflows within ODH so the trained model can be packaged in a container, tested and pushed from within ODH

OK. So is the plan to iterate in this repository on those integrations?

> ODH 1.x is planned for deprecation with the 1.11 release (See [opendatahub-operator #648](https://github.com/opendatahub-io/opendatahub-operator/issues/648)). The `alpha` reference in opendatahub.io is just an oversight as it was used as a preview for the downstream release that started in July. We accelerated the development of v2.x and the opendatahub.io documentation for the ODH install just needs a refresh

Thanks for that https://github.com/opendatahub-io/opendatahub-operator/issues/648 link, that sums it nicely.

LaVLaS:
Closing this to follow-up on with another PR to update the documentation across the board

PR #176, opened 2023-11-22
feat: can now run basic tests on the cluster

jackdelahunt:
## Merge criteria:


test/e2e-tests/support/client_set.go
```diff
@@ -0,0 +1,33 @@
...
+	"path/filepath"
+)
+
+func ClusterClientSet() (*kubernetes.Clientset, error) {
+	var kubeConfig *string
+	if home := homedir.HomeDir(); home != "" {
+		kubeConfig = flag.String("kubeconfig", filepath.Join(home, ".kube", "config"), "(optional) absolute path to the kubeconfig file")
+	} else {
+		kubeConfig = flag.String("kubeconfig", "", "absolute path to the kubeconfig file")
+	}
```
adelton, line 18: Could we just use `KUBECONFIG` environment variable here, and if needed, set the default in the Makefile?

test/e2e-tests/Makefile
```diff
@@ -0,0 +1,3 @@
+.PHONY: test
+test:
+	@(cd tests && go test)
```
adelton, line 3: For people who might not (want to) have golang installed on their machines, could you also show a target which would run the tests from a local container?

test/e2e-tests/support/client_set.go
jackdelahunt: Is the `KUBECONFIG`enviroment variable set automatically for you with `kubectl` or `oc` atleast for me it isnt set when I log in. And setting a default in the Makefile doesn't seem right to me, I would rather all config be handled through arguments rather than enviroment variables set by a Makefile

test/e2e-tests/Makefile
jackdelahunt: I could but is this really a need? We are running tests for a k8s cluster locally I think expecting go to be installed is okay. Especially because if you have docker or podman installed on your machine why would go be a deal breaker?

adelton: I have rootless podman setup exactly because most everything (including for example `oc` or `rosa`) is run in containers. It's because being involved with multiple projects, the requirements for the versions of compilers and libraries (not just golang) that I'd have to match seem to differ wildly and in conflicting manner, so the most sane way is to just have it all containerized.

Even K8s itself does not require you to have golang installed to build it -- when following https://github.com/kubernetes/kubernetes#to-start-developing-k8s the `make quick-release` works through docker. And with https://github.com/kubernetes/kubernetes/pull/121533 even with podman.

test/e2e-tests/support/client_set.go
grdryn: > Is the `KUBECONFIG` enviroment variable set automatically for you with `kubectl` or `oc` atleast for me it isnt set when I log in.

No, it's an alternative where you can point to a different kubeconfig file for different clusters/contexts, rather than needing to interactively login/switch between clusters/contexts that are specified in a single file in the default location (`~/.kube/config`).

It might be more common of a pattern in the vanilla k8s world than OpenShift, since I don't think there's even a `login` subcommand on the `kubectl` CLI (at least not on the version that I've got, which might be a little old).

I'm +1 for using `KUBECONFIG` since it's a known good approach, and potentially friendlier for running in environments like containers or CI.

jackdelahunt: Okay I have updated the PR to do as follows:
- check if the `KUBECONFIG` env variable is set and if so it uses that location
- Else it just uses the default location `~/.kube/config`

test/e2e-tests/Makefile
LaVLaS: If we are standardizing around `gotest` for the test framework it would imply that `go` is available locally.

@jackdelahunt One recommendation I would make is that you replace the call to `go` with a Makefile variable that defaults to `GO = go`.  This would add support for calling a specific`goX.Y.Z` binary version or absolute path 
```
make GO=go1.19 test
```

@adelton I use the same workflow for my local development.  For that use case, I would redirect anyone to [toolbox](https://github.com/containers/toolbox) or [distrobox](https://github.com/89luca89/distrobox) if they want to run those binaries from a containerized environment with the current directory mounted

LaVLaS: ...and can you add a new line to the end of the file

test/e2e-tests/README.md
```diff
@@ -0,0 +1,11 @@
+# AI Edge testing
+
+Place for testing and to verify the MLOps pipelines are working correctly. These tests use the `k8s.io/client-go` package to interact with the cluster. Using `oc login` to log into the cluster as normal should mean no setup is needed to run the tests.
```
LaVLaS, line 3: Can you add a mention of the minimum go version required for these tests?

test/e2e-tests/Makefile
jackdelahunt: Sure sounds good to me 👍

test/e2e-tests/support/client_set.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"k8s.io/client-go/kubernetes"
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/util/homedir"
+	"os"
+	"path/filepath"
+)
+
+func ClusterClientSet() (*kubernetes.Clientset, error) {
```
winklerm, line 12: I am thinking how this will work for two clusters, each having a different KubeConfig, that a test developer would like to use both in a single test? 

Maybe there could be a variant of this function with a single parameter allowing to specify the KubeConfig location (or the name of the env. var to use instead of `KUBECONFIG`)?

On the other hand, this is a nice start and it can be extended later when we know more what is needed regarding two cluster support.

winklerm:
@MarianMacik FYI

test/e2e-tests/support/client_set.go
jackdelahunt: Would this be something we are looking for? Because if we are switching contexts then only one kubeconfig should be needed right? You just need to log into them both

LaVLaS: > On the other hand, this is a nice start and it can be extended later when we know more what is needed regarding two cluster support.

+1 to this.  Right now the existing functionality is focused on providing the core of a go test framework to interact with an kubernetes cluster.  As we plan new features for this repo then we can design the test framework in parallel.

MarianMacik: Yes, there are basically two options:
* switch contexts
* 2 kubeconfig files
We will need to find out what is more convenient. It will depend on the output of the provisioning job, when it outputs 2 user/password pairs, we would probably use one kubeconfig. If it outputs 2 kubeconfig files, it might be easier for us to reuse these 2 kubeconfig files.

Wdyt @winklerm?

MarianMacik (approved):

test/e2e-tests/support/client_set.go
jackdelahunt: > +1 to this. Right now the existing functionality is focused on providing the core of a go test framework 

Sounds good

test/e2e-tests/Makefile
```diff
@@ -0,0 +1,5 @@
+GO=go
+
+.PHONY: test
+test:
+	@(cd tests && ${GO} test)
```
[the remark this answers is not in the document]
jackdelahunt: @LaVLaS in the comment for this dummy make target it says it is for openshift-ci. So is this not meant for the go tests then?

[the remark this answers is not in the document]
jackdelahunt: Okay moved the target to run from the e2e-test folder to the root as yous said. Also updated the README to describe how to run the tests

jackdelahunt:
I moved the make target from the tests sub-directory to the root. Openshift-ci runs this but because we don't have it setup (and this PR is not supposed to address that) it fails. This failing means the PR can not be merged.

So we need to wait until we actively try and make these tests work in CI to have the tests runnable from the root.

@MarianMacik

LaVLaS:
> I moved the make target from the tests sub-directory to the root. Openshift-ci runs this but because we don't have it setup (and this PR is not supposed to address that) it fails. This failing means the PR can not be merged.

This is probably the standard chicken or egg scenario when updating core functionality that is referenced by openshift-ci.  For review of this PR, if we can run `make test` from the root of the repo and it kicks off `tests/e2e-test` I would consider it LGTM.  WYD @MarianMacik @winklerm

MarianMacik:
@LaVLaS @winklerm @jackdelahunt The OpenShift CI is correctly set up. The go test command was run but it failed with:
```
go: inconsistent vendoring in /go/src/github.com/opendatahub-io/ai-edge/test/e2e-tests:
	k8s.io/apimachinery@v0.28.4: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	k8s.io/client-go@v0.28.4: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/davecgh/go-spew@v1.1.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt 
...
```
So this is not an issue connected to an OCP cluster or something provisioned by OpenShift CI. However, there is not any vendor directory as we use only modules.
I guess the issue is probably that we use golang 1.19 image on OpenShift CI, I will update it to 1.21. Maybe it's incompatible and this is a side effect.

jackdelahunt:
@MarianMacik I wasn't able to see the logs thanks for showing me. Maybe we just need to run `go vendor` to add the dependencies to the vendor folder first then run go test?

MarianMacik:
@jackdelahunt AFAIK the vendor mode was used before the Go modules were implemented and now it shouldn't be created at all.

jackdelahunt:
No I thought vendoring was a seperate thing all together, where instead of installing the deps in the go folder for your system and sharing them across projects it stores them per project in the vendor folder to be used in source control. This is so you can not rely on the network/go fethcing the deps, to build the project. 

Before I thought you had modules you needed to build the project in the go folder, and deps where used from the global go folder.

LaVLaS:
>  This is so you can not rely on the network/go fethcing the deps, to build the project.
+1 to this for `go mod vendor`. It's present day use case is to embed a snapshot of your module versions in your repo to ensure that they never break due to any change in one dependency.

I think the openshift-ci system is setting `GOFLAGS=-mod=vendor` which is forcing the `vendor` directory.  We can try to unset `GOFLAGS=` in the openshift-ci run OR explicitly set `GOFLAGS=-mod=mod` specifically for the CI tests.  
See [go build commands](https://go.dev/ref/mod#build-commands)

jackdelahunt:
Okay I created the `vendor` directory for the testing suite and explicity set the `GOFLAGS` to use vendoring @LaVLaS

LaVLaS:
> Okay I created the vendor directory for the testing suite and explicitly set the GOFLAGS to use vendoring @LaVLaS

just to clarify. The openshift-ci system is  setting `GOFLAGS=-mod=vendor` which our execution of `make test` is inheriting. 


We need to unset it explicitly in openshift-ci test run so I submitted https://github.com/openshift/release/pull/46372 to openshift-ci to resolve the issue.
 
 @jackdelahunt You can remove the `vendor` directory because it is only needed in very specific use cases where a high degree of reproducibility and stability for each release is absolutely required.  Once https://github.com/openshift/release/pull/46372, you can retest the ability to run `make test` from the repo root to ensure everything is working as intended

test/e2e-tests/README.md
```diff
@@ -0,0 +1,17 @@
...
+```bash
+make test
+```
+Set the go binary used for testing that is in your `PATH`
+```bash
+make GO=go1.19 test
+```
+
+## CI/CD with Github Actions
+Not yet implemented
```
grdryn, line 17: I guess it's OpenShift CI rather than GitHub Actions now, so these lines may not be relevant any longer?

test/e2e-tests/README.md
```diff
@@ -0,0 +1,17 @@
+# AI Edge testing
+
+Place for testing and to verify the MLOps pipelines are working correctly. These tests use the `k8s.io/client-go` package to interact with the cluster. Using `oc login` to log into the cluster as normal should mean no setup is needed to run the tests.
+
+A locally install of the Go compiler is needed to run the tests. Go version `1.21` is required to run the tests.
```
grdryn, line 5:
```suggestion
A local install of the Go compiler is needed to run the tests. Go version `1.21` is required to run the tests.
```

grdryn: Per this comment, is Go 1.21 what we should be using? It looks like we've configured Prow to use 1.19: https://github.com/openshift/release/blob/master/ci-operator/config/opendatahub-io/ai-edge/opendatahub-io-ai-edge-main.yaml#L5

@LaVLaS ^^

LaVLaS: We can definitely update the openshift-ci version if it causes issues.  I assume we just pinned the current release to 1.21 since it was the recent version and I don't think there are major changes that would affect running go test in go <= v1.19

test/e2e-tests/README.md
LaVLaS: This will probably be a dual ownership for both CI/CD workflows.  For github actions, we can execute of free workflows to linting and running some lightweight workflows in a [kind](https://kind.sigs.k8s.io/) cluster.  openshift-ci would be utilized for larger integration testing

test/e2e-tests/README.md
LaVLaS: @jackdelahunt Everything looks good.  The suggestion above is the last [minor] issue to resolve and we can merge

test/e2e-tests/README.md
grdryn: Ah right, understood, thanks :+1:

test/e2e-tests/README.md
jackdelahunt: 👍

grdryn (approved):
/lgtm

PR #188, opened 2023-12-11
RHOAIENG-201: Add e2e-tests to verify the MLOps pipelines are working correctly

jackdelahunt:
## Description
Adds two make targets `go-test` `go-test-setup`. Running `make go-test` with its required environment variables defined will begin the tests. A working namespace `test-namespace` is required to for both the setup and test target

## How Has This Been Tested?
- Log into a cluster with `oc`
- Set the list of enviroment variables that are needed for running the tests for both the `go-test` and `go-test-setup` make targets. All of the required fields are listed [here](https://github.com/jackdelahunt/ai-edge/blob/pipeline-testing/Makefile#L37-L44) and [here](https://github.com/jackdelahunt/ai-edge/blob/pipeline-testing/Makefile#L74-L76)
- Run `oc new-project test-namespace`
- Run `make AWS_SECRET=... AWS_ACCESS=...  ... go-test-setup`
- Run `make S3_BUCKET=...  ... go-test`
- All tests should pass

### Note
- Setting `GO` to the path to any go compiler you may have when running `go-test` will overwrite the compiler used.

## Merge criteria:


test/e2e-tests/tests/sanity_test.go
```diff
@@ -1,19 +1,68 @@
...
+	if err != nil {
+		t.Fatal(err.Error())
+	}
+
+	ctx := support.CreateContext()
+
+	if false {
+		support.CreateTestNamespace(ctx, clientSet)
+		defer support.DeleteTestNamespace(ctx, clientSet)
+	}
```
jackdelahunt, line 32: While these create and delete the namespace as expected. Using `oc. project` to set them as the current namespace for all other oc commands does not work. Whatever project you were using before testing will be used.

Deleting the namespace does not wait for it to be deleted. Which means calling multiple tests with this will cause an error because you are trying to create and delete the namespace at the same time

test/e2e-tests/tests/sanity_test.go
```diff
@@ -1,19 +1,68 @@
...
+	"github.com/opendatahub-io/ai-edge/test/e2e-tests/oc"
 	support "github.com/opendatahub-io/ai-edge/test/e2e-tests/support"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"testing"
 )
 
-func TestGetAllPods(t *testing.T) {
+const (
+	ContainerImagePipeline    = "../../../pipelines/tekton/build-container-image-pipeline/"
+	ContainerImagePipelineRun = "../../../pipelines/tekton/build-container-image-pipeline/build-container-image-pipelinerun-bike-rentals.yaml"
```
jackdelahunt, line 11: This works but of course it is not ideal. I am not really sure how else we want to specify where out pipelines and pipelineRuns are located. 

Maybe though an envriment variable which is set the Makefile? I am not sure

jackdelahunt:
I am not sure how we can check the progress of a currently running pipelineRun, so right now that is not here

LaVLaS:
@jackdelahunt Can you add a readme to the `tests/` to provide a quick summary of what this test framework covers and how to execute it?

jackdelahunt:
> @jackdelahunt Can you add a readme to the `tests/` to provide a quick summary of what this test framework covers and how to execute it?

I already have a README in the `test/e2e-tests` is that one okay?

test/e2e-tests/oc/oc.go
```diff
@@ -0,0 +1,56 @@
+package oc
```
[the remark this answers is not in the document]
jackdelahunt: I thought we mentioned that we would just use `oc` directly.  But I did also read into it, pushing non-standard k8s object is a bit involved so I kept to what we said. But if you want I can update the PR to just use the `client-go` package.

LaVLaS:
> I already have a README in the `test/e2e-tests` is that one okay?

That one is fine.  Once this is ready for final review, we should add some info for the the initialization and execution flow

jackdelahunt:
> That one is fine. Once this is ready for final review
I wouldn't say this is ready for final review. I still have questions about what we are doing and also have some issues I think.

Really about the comments above.

jackdelahunt:
Prow is failing because the tests are timing out after 10m. The tests shouldn't take this long so there are issues with the tests themselves. I have started to talk to @LaVLaS about this issue.

Makefile
```diff
@@ -34,4 +34,8 @@ GOFLAGS=""
 test:
 	@(./test/shell-pipeline-tests/openvino-bike-rentals/pipelines-test-openvino-bike-rentals.sh)
 	@(./test/shell-pipeline-tests/tensorflow-housing/pipelines-test-tensorflow-housing.sh)
+
+	@(oc create namespace test-namespace && oc project test-namespace)
```
LaVLaS, line 38: Can you separate this into a separate target for running `go test`? Until we can refactor the and unify the setup for the `shell` and `go test` framework, we'll need to have separate targets to prevent CI/CD runs from failing when both are not aligned

jackdelahunt:
So far what is needed is passing the bucket name and image repo name. Right now the go test needs these for changing in the pipelineRun.

test/e2e-tests/tests/pipelines_test.go
```diff
@@ -0,0 +1,156 @@
...
+		t.Fatalf("expected %v pipelines but %v existed on the cluster", len(expectedPipelineNames), len(list.Items))
+	}
+
+OUTER:
+	for _, name := range expectedPipelineNames {
+		for _, pipeline := range list.Items {
+			if pipeline.Name == name {
+				continue OUTER
+			}
+		}
```
MarianMacik, line 104: This can be extracted to a function which would return true if the pipeline was found and one if, which would check the outcome. This way the label can be omitted. But again, it may be just me writing it differently, no strong opinion :)

test/e2e-tests/tests/pipelines_test.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	err = WaitForAllPipelineRunsToComplete(ctx, &clients)
+	if err != nil {
+		t.Fatal(err.Error())
+	}
+}
+
+func Test_PipelineRunsCompleted(t *testing.T) {
+	t.Skip()
```
MarianMacik, line 144: I guess this is because of work in progress?

MarianMacik (commented):
Looks good in general. I have just suggested a few things, you don't need to apply all of them if you find the current code is OK.

test/e2e-tests/support/client_set.go
```diff
@@ -1,14 +1,49 @@
...
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/util/homedir"
+	knativetest "knative.dev/pkg/test"
 	"os"
 	"path/filepath"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+type Clients struct {
+	ControllerRuntime client.Client
```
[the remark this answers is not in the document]
jackdelahunt: Yep I just left that in there by mistake was trying to use it but could't get the connection to the cluster to work. Looking at the `client.Create()` that is what I wanted instead of using each client.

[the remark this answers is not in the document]
jackdelahunt: Removed.. for now atleast 👍

test/e2e-tests/support/client_set.go
```diff
@@ -1,14 +1,49 @@
...
+
+	clientSet, err := ClusterClientSet()
+	if err != nil {
+		return clients, err
+	}
+
+	config, err := knativetest.BuildClientConfig(knativetest.Flags.Kubeconfig, knativetest.Flags.Cluster)
+	if err != nil {
+		return clients, err
+	}
```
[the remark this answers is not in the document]
jackdelahunt: Done 👍

test/e2e-tests/support/options.go
```diff
@@ -0,0 +1,55 @@
...
+		return options, nil
+	}
+
+	var options = &Options{}
+
+	if os.Getenv(S3BucketNameEnvKey) != "" {
+		options.S3BucketName = os.Getenv(S3BucketNameEnvKey)
+	} else {
+		return options, fmt.Errorf("env variable %v not set, but is required to run tests", S3BucketNameEnvKey)
+	}
```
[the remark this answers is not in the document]
jackdelahunt: I didn't know you could do that, nice 👍

test/e2e-tests/support/options.go
```diff
@@ -0,0 +1,55 @@
...
+
+	if os.Getenv(RegistryRepoNameEnvKey) != "" {
+		options.RegistryRepoName = os.Getenv(RegistryRepoNameEnvKey)
+	} else {
+		return options, fmt.Errorf("env variable %v not set, but is required to run tests", RegistryRepoNameEnvKey)
+	}
```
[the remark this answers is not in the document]
jackdelahunt: Done 👍

Makefile
```diff
@@ -34,7 +34,66 @@ endif
...
+	@cat pipelines/tekton/aiedge-e2e/templates/credentials-s3.secret.yaml.template | sed 's#AWS_SECRET#${AWS_SECRET}#g' | sed 's#AWS_ACCESS#${AWS_ACCESS}#g' | sed 's#S3_REGION#${S3_REGION}#g' | sed 's#S3_ENDPOINT#${S3_ENDPOINT}#g' | oc create -f -
+	@cat pipelines/tekton/aiedge-e2e/templates/credentials-image-registry.yaml.template | sed 's#REGISTRY_USERNAME#${REGISTRY_USERNAME}#g' | sed 's#REGISTRY_PASSWORD#${REGISTRY_PASSWORD}#g' | oc create -f -
+	oc create -f ${REGISTRY_CREDENTIALS}
+	oc secret link pipeline $(shell yq '.metadata.name' ${REGISTRY_CREDENTIALS})
+	oc create -f pipelines/tekton/aiedge-e2e/test-data/bike-rentals-test-data-cm.yaml
+
+# requires:
+#	S3_BUCKET		- Name of S3 bucket that has the model
+#	REGISTRY_REPO		- quay.io robot secret file
+go-test:
```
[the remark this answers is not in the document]
jackdelahunt: Done 👍

test/e2e-tests/tests/pipelines_test.go
jackdelahunt: I actually just removed the extra tests, nether were needed any more and or did the same as the current test

test/e2e-tests/support/tekton.go
```diff
@@ -0,0 +1,69 @@
...
+	if err != nil {
+		return pipelineRun, err
+	}
+
+	// Read the file into a byte slice
+	bytes := make([]byte, stat.Size())
+	_, err = bufio.NewReader(file).Read(bytes)
+	if err != nil && err != io.EOF {
+		return pipelineRun, err
+	}
```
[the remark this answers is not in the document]
jackdelahunt: Yep done 👍

test/e2e-tests/tests/pipelines_test.go
jackdelahunt: Deleted

test/e2e-tests/tests/pipelines_test.go
```diff
@@ -0,0 +1,156 @@
...
+		list, err := clients.PipelineRun.List(ctx, k8sApi.ListOptions{})
+		if err != nil {
+			return err
+		}
+
+		for _, pipelineRun := range list.Items {
+			if len(pipelineRun.Status.Conditions) == 0 {
+				time.Sleep(time.Second * 5)
+				continue CheckRuns // waiting for conditions to exist first
+			}
```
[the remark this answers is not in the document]
jackdelahunt: Okay I updated it, don't think its great still because of using two bools, but you can let me know what you think 
👍

test/e2e-tests/support/options.go
```diff
@@ -0,0 +1,51 @@
...
+	"os"
+)
+
+const (
+	S3BucketNameEnvKey     = "S3_BUCKET"
+	RegistryRepoNameEnvKey = "REGISTRY_REPO"
+)
+
+var (
+	options *Options = nil
```
grdryn, line 14: I have some questions about this, after looking at how it's used:

- Options is quite a generic name. Is this expected to grow to hold all potential things that could be passed in or collected from the environment and used in the tests, or just pipeline run parameter values?
- This looks like an shared global variable between all tests: if we want to execute different PipelineRuns (either in the same test or in different tests), are we limited to using, for example, the same `target-imagerepo` for each?

Might not be worth thinking or doing anything about these questions until we need to, just thinking out loud.

grdryn (commented):
Some comments inline (some may not be valid), but general approach seems reasonable. :+1:

Makefile
```diff
@@ -34,7 +34,57 @@ endif
...
+	@cat pipelines/tekton/aiedge-e2e/templates/credentials-s3.secret.yaml.template | sed 's#{{ AWS_SECRET }}#${AWS_SECRET}#g' | sed 's#{{ AWS_ACCESS }}#${AWS_ACCESS}#g' | sed 's#{{ S3_REGION }}#${S3_REGION}#g' | sed 's#{{ S3_ENDPOINT }}#${S3_ENDPOINT}#g' | oc create -f -
+	@cat pipelines/tekton/aiedge-e2e/templates/credentials-image-registry.yaml.template | sed 's#{{ REGISTRY_USERNAME }}#${REGISTRY_USERNAME}#g' | sed 's#{{ REGISTRY_PASSWORD }}#${REGISTRY_PASSWORD}#g' | oc create -f -
+	oc create -f ${REGISTRY_CREDENTIALS}
+	oc secret link pipeline $(shell yq '.metadata.name' ${REGISTRY_CREDENTIALS})
+	oc create -f pipelines/tekton/aiedge-e2e/test-data/bike-rentals-test-data-cm.yaml
+
+# requires:
+#	S3_BUCKET		- Name of S3 bucket that has the model
+#	REGISTRY_REPO		- quay.io robot secret file
+go-test go-test-setup:
```
MarianMacik, line 77: I believe prerequisite should be written like this:
```suggestion
go-test: go-test-setup
```

Makefile
```diff
@@ -34,7 +34,57 @@ endif
...
 GOFLAGS=""
 
+# requires:
+#	AWS_SECRET		- Secret from AWS
+#	AWS_ACCESS		- Access key from AWS
+#	S3_REGION		- Region of the bucket used to store the model
+#	S3_ENDPOINT		- Endpint of the bucket
+#	REGISTRY_USERNAME	- quay.io username
+#	REGISTRY_PASSWORD	- quay.io password
+#	REGISTRY_CREDENTIALS	- quay.io robot secret file
```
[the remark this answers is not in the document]
MarianMacik: To be honest this seems to be used just in `oc create -f ${REGISTRY_CREDENTIALS}` but that would create unsubstituted secret.

Makefile
```diff
@@ -34,7 +34,57 @@ endif
...
+endif
+ifndef REGISTRY_PASSWORD
+	$(error REGISTRY_PASSWORD is undefined)
+endif
+ifndef REGISTRY_CREDENTIALS
+	$(error REGISTRY_CREDENTIALS is undefined)
+endif
+	oc new-project test-namespace
+	@cat pipelines/tekton/aiedge-e2e/templates/credentials-s3.secret.yaml.template | sed 's#{{ AWS_SECRET }}#${AWS_SECRET}#g' | sed 's#{{ AWS_ACCESS }}#${AWS_ACCESS}#g' | sed 's#{{ S3_REGION }}#${S3_REGION}#g' | sed 's#{{ S3_ENDPOINT }}#${S3_ENDPOINT}#g' | oc create -f -
+	@cat pipelines/tekton/aiedge-e2e/templates/credentials-image-registry.yaml.template | sed 's#{{ REGISTRY_USERNAME }}#${REGISTRY_USERNAME}#g' | sed 's#{{ REGISTRY_PASSWORD }}#${REGISTRY_PASSWORD}#g' | oc create -f -
```
MarianMacik, line 69: Also, the secret changed name to `credentials-image-registry.secret.yaml.template`.
```suggestion
	@cat pipelines/tekton/aiedge-e2e/templates/credentials-image-registry.secret.yaml.template | sed 's#{{ REGISTRY_USERNAME }}#${REGISTRY_USERNAME}#g' | sed 's#{{ REGISTRY_PASSWORD }}#${REGISTRY_PASSWORD}#g' | oc create -f -
```

pipelines/tekton/aiedge-e2e/templates/credentials-s3.secret.yaml.template
```diff
@@ -8,4 +8,4 @@ metadata:
 
 stringData:
   s3-storage-config: |+
-    { "type": "s3", "access_key_id": "{{ YOUR_AWS_ACCESS_KEY }}", "secret_access_key": "{{ YOUR_AWS_SECRET_KEY }}", "endpoint_url": "{{ S3_ENDPOINT__https://example.amazonaws.com/ }}", "region": "{{ S3_REGION__us-west-1 }}" }
+    { "type": "s3", "access_key_id": "{{AWS_ACCESS}}", "secret_access_key": "{{AWS_SECRET}}", "endpoint_url": "{{S3_ENDPOINT}}", "region": "{{S3_REGION}}" }
```
MarianMacik, line 11: This will need a change in the shell tests, otherwise the PR check will fail.

test/e2e-tests/support/client_set.go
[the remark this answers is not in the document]
MarianMacik: Yeah, in the future we should either use the controller-runtime client or the dynamic client. It wasn't probably working for you because you didn't configure schemas for that connection? We can have a look at it in the future.

test/e2e-tests/tests/pipelines_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+func CreateContext() context.Context {
+	return context.Background()
+}
+
+func WaitForAllPipelineRunsToComplete(ctx context.Context, clients *support.Clients) error {
+	checkPipelineRuns := true
+
+	for checkPipelineRuns {
+		time.Sleep(time.Second * 5)
```
MarianMacik, line 28: Here I meant it a little bit different. The wait loop for conditions to be present was OK, the loop just could wait until the `len(pipelineRun.Status.Conditions) == 0` is false.



So to sum up I was thinking of something like:

```go
list, err := clients.PipelineRun.List(ctx, k8sApi.ListOptions{})

For each pipeline from the list {
    Wait for the conditions to appear in a for loop{}
    pipelineRunning := true
    For pipelineRunning {
        For each condition {
            If Failed, return error, elseif Running wait 5 seconds, else pipelineRunning = false // or similar
        }
    }
}
```
I believe this would work the same. We can wait on a specific pipeline because we anyway want all of them to complete. So we don't need to check all of them periodically, we just need to finish when the last pipeline finishes. Unless the pipelines can be started while this check is already running, in that case you may want to fetch "fresh" pipeline list every time. But here the usual use case will be run all pipelines and then wait.

Makefile
```diff
@@ -34,7 +34,57 @@ endif
...
+#	REGISTRY_USERNAME	- quay.io username
+#	REGISTRY_PASSWORD	- quay.io password
+#	REGISTRY_CREDENTIALS	- quay.io robot secret file
+go-test-setup:
+ifndef AWS_SECRET
+	$(error AWS_SECRET is undefined)
+endif
+ifndef AWS_ACCESS
+	$(error AWS_ACCESS is undefined)
+endif
```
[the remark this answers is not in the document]
LaVLaS: Agreed. Those have always been the standard names we have used across ODH workshops and tutorials since it was o[riginal ly support by the AWS SDK](https://docs.aws.amazon.com/sdkref/latest/guide/settings-reference.html#EVarSettings) so all S3 compatible APIs and will provide some level of support for those names

Makefile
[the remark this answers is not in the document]
LaVLaS: I think we should simplify this entire setup workflow to have the user provide their own unique registry & credentials files instead of using the Makefile to process them.  

Both files are a known format and one time initialization and we only need the secretName for the PipelineRun execution.  This should simplify the workflow of having to setup the environment with secure information

Makefile
```diff
@@ -34,7 +34,57 @@ endif
...
+	oc new-project test-namespace
+	@cat pipelines/tekton/aiedge-e2e/templates/credentials-s3.secret.yaml.template | sed 's#{{ AWS_SECRET }}#${AWS_SECRET}#g' | sed 's#{{ AWS_ACCESS }}#${AWS_ACCESS}#g' | sed 's#{{ S3_REGION }}#${S3_REGION}#g' | sed 's#{{ S3_ENDPOINT }}#${S3_ENDPOINT}#g' | oc create -f -
+	@cat pipelines/tekton/aiedge-e2e/templates/credentials-image-registry.yaml.template | sed 's#{{ REGISTRY_USERNAME }}#${REGISTRY_USERNAME}#g' | sed 's#{{ REGISTRY_PASSWORD }}#${REGISTRY_PASSWORD}#g' | oc create -f -
+	oc create -f ${REGISTRY_CREDENTIALS}
+	oc secret link pipeline $(shell yq '.metadata.name' ${REGISTRY_CREDENTIALS})
+	oc create -f pipelines/tekton/aiedge-e2e/test-data/bike-rentals-test-data-cm.yaml
+
+# requires:
+#	S3_BUCKET		- Name of S3 bucket that has the model
+#	REGISTRY_REPO		- quay.io robot secret file
```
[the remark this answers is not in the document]
jackdelahunt: Copy paste error 👍

test/e2e-tests/support/kustomize.go
```diff
@@ -0,0 +1,101 @@
...
+
+	err = json.Unmarshal(bytes, &t)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func ApplyPipelineKustomizeFolder(path string) (resmap.ResMap, error) {
```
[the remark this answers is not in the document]
jackdelahunt: Renamed to `KustomizeBuild`

test/e2e-tests/support/tekton.go
```diff
@@ -0,0 +1,53 @@
...
+	var pipelineRun v1.PipelineRun
+
+	bytes, err := os.ReadFile(path)
+	if err != nil {
+		return pipelineRun, err
+	}
+
+	err = yaml.Unmarshal(bytes, &pipelineRun)
+	if err != nil {
+		return pipelineRun, err
```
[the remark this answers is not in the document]
jackdelahunt: >  what will the value of pipelineRun be
Because in these cases `err != nil` when the caller checks the return values from this function then there is no expectation for the value of pipelineRun

> Should nil be explicitly returned instead?
Because I am returning by value some pipelineRun value does need to be returned. I could return a default initialised one like `PipelineRun{}` or do what I did here and just return the value we currently have 

What do you think?

test/e2e-tests/tests/pipelines_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"time"
+
+	"github.com/opendatahub-io/ai-edge/test/e2e-tests/support"
+	"golang.org/x/net/context"
+	k8sApi "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	PipelineRelativePath    = "../../../pipelines/tekton/aiedge-e2e"
+	PipelineRunRelativePath = "../../../pipelines/tekton/aiedge-e2e/aiedge-e2e.pipelinerun.yaml"
```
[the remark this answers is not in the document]
jackdelahunt: Done 👍

test/e2e-tests/tests/pipelines_test.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	ctx := CreateContext()
+
+	clients, err := support.CreateClients("test-namespace")
+	if err != nil {
+		panic(fmt.Sprintf("error while creating client : %v", err.Error()))
+	}
+
+	for _, resource := range resourceMap.Resources() {
```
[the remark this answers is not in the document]
jackdelahunt: Done 👍

test/e2e-tests/support/tekton.go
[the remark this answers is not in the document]
grdryn: Ah yes, understood. :+1: I _guess_ the value of pipelineRun is `PipelineRun{}` at this point then?

pipelines/tekton/aiedge-e2e/templates/credentials-s3.secret.yaml.template
jackdelahunt: I made this change in this comment, tell me it is wrong because maybe there is more 

https://github.com/opendatahub-io/ai-edge/pull/188/commits/2ed4f2b2982def7295feb8ca29217f6aa35d9bf7

test/e2e-tests/tests/pipelines_test.go
jackdelahunt: I am kind of confused, but maybe I am just not understanding. In the example above are we checking the same conditions each time, and if so what would make that condition change for us to get out of the `pipelineRunning` loop. I thought we would need to call `PipelineRun.List` to get the up to date info on the condition of each pipelineRun.

test/e2e-tests/support/options.go
jackdelahunt: > Options is quite a generic name. Is this expected to grow to hold all potential things that could be passed in or collected from the environment and used in the tests, or just pipeline run parameter values?

My thinking was any options that can be used by the tests which are set by environment variables.

> This looks like an shared global variable between all tests: if we want to execute different PipelineRuns 

They are only shared if a tests wants to use them. Right now the structure I am thinking about and that I have talked to landon about is each test constructs one or more pipelineRuns and we see if it is a success.

If a tests does not want to use the options passed, it can just set that for a different value in the tests. The same way I am doing it https://github.com/opendatahub-io/ai-edge/pull/188/files#diff-a0b087c51d28b57a9b15240b642f35b5425c1bd662312f637ba56e176ecde3e1R92-R93

jackdelahunt: Also Options is generic but that is how I was thinking of them, the same as arguments but instead passing them as args they are just set and env variables. If we want to change it that is fine with me. Maybe `TestingArguments`?

Makefile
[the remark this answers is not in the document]
jackdelahunt: > It seems odd to require username and password, and also separately a secret file for "credentials"

True, the credentials file is for the quay robot you use to push to your repo. Either way we would need to set them in the file so I think we need to think about these sepretely at some point anyway, either as test arguments or before when setting up the credentials file

> To be honest this seems to be used just in oc create -f ${REGISTRY_CREDENTIALS} but that would create unsubstituted secret

I am not really sure what this means, should I not be doing it this way? @MarianMacik

test/e2e-tests/tests/pipelines_test.go
MarianMacik: @jackdelahunt Ah, sorry, I didn't consider that, my bad. However, after fetching the initial pipeline list you can always fetch a single fresh pipeline run before each condition check. This way the loop would be probably cleaner. Also you can merge waiting for the condition to appear loop with the pipelineRunning loop. E.g.:

```go
list, err := clients.PipelineRun.List(ctx, k8sApi.ListOptions{})

For each pipeline from the list {
    pipelineRunning := true
    For pipelineRunning {
        // Here fetch a single pipeline run and get its conditions
        // If no condition present, wait 5 seconds (the next loop will be noop) and let the pipelineRunning loop execute again or you can explicitly use continue
        For each condition {
            If Failed, return error, elseif Running wait 5 seconds, else pipelineRunning = false // or similar
        }
    }
}
```

I think this is cleaner than what was before but this is just my suggestion.

test/e2e-tests/tests/pipelines_test.go
```diff
@@ -0,0 +1,108 @@
...
+	PipelineRunRelativePath = "../../../pipelines/tekton/aiedge-e2e/aiedge-e2e.pipelinerun.yaml"
+	S3BucketNameEnvKey      = "S3_BUCKET"
+	RegistryRepoNameEnvKey  = "REGISTRY_REPO"
+)
+
+func CreateContext() context.Context {
+	return context.Background()
+}
+
+func WaitForAllPipelineRunsToComplete(ctx context.Context, clients *support.Clients) error {
```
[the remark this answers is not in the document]
grdryn: That's probably sufficient for now. If we need a version of this with a timeout in future, we can wrap this function. :+1:

test/e2e-tests/support/options.go
grdryn: I think this is fine for now. I wouldn't like to see it become a huge object that has lots of configurable knobs through environment variables. Let's just try to be mindful of what we might add to it, and not be afraid to use another approach if it suits us better in future :+1:

test/e2e-tests/support/kustomize.go
```diff
@@ -0,0 +1,101 @@
...
+		return nil, err
+	}
+
+	return resourceMap, nil
+}
+
+func CreateObjectFromResource(ctx context.Context, clients *Clients, resource *resource.Resource) error {
+		kind := resource.GetKind()
+
+		switch kind {
```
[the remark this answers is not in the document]
jackdelahunt: I added the default case just to return an error if we ever add other kinds of objects in the build. I kept the assignment of kind outside of the switch as I think it is a bit more readable.

test/e2e-tests/tests/pipelines_test.go
jackdelahunt: I did as you said and I am not really sure if it ended up looking cleaner but you can take a look

Makefile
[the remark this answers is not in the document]
jackdelahunt: I did add this change but I am not sure on Landon's comment

test/shell-pipeline-tests/common.sh
```diff
@@ -51,10 +51,10 @@ function createS3Secret() {
...
     cp "$AWS_SECRET_PATH_TEMPLATE" "$AWS_SECRET_PATH"
 
-    sed -i "s|{{ YOUR_AWS_ACCESS_KEY }}|${AWS_ACCESS_KEY}|" "$AWS_SECRET_PATH"
-    sed -i "s|{{ YOUR_AWS_SECRET_KEY }}|${AWS_SECRET_KEY}|" "$AWS_SECRET_PATH"
-    sed -i "s|{{ S3_ENDPOINT__https://example.amazonaws.com/ }}|https://s3.us-west-1.amazonaws.com|" "$AWS_SECRET_PATH"
-    sed -i "s|{{ S3_REGION__us-west-1 }}|us-west-1|" "$AWS_SECRET_PATH"
+    sed -i "s|{{ AWS_ACCESS_KEY_ID }}|${AWS_ACCESS_KEY}|" "$AWS_SECRET_PATH"
+    sed -i "s|{{ AWS_SECRET_ACCESS_KEY }}|${AWS_SECRET_KEY}|" "$AWS_SECRET_PATH"
+    sed -i "s|{{ S3_ENDPOINT }}|" "$AWS_SECRET_PATH"
+    sed -i "s|{{ S3_REGION}}|" "$AWS_SECRET_PATH"
```
MarianMacik, line 57: Endpoint and region are now not substituted.

test/e2e-tests/tests/pipelines_test.go
MarianMacik: Looks good to me, however, [this one](https://github.com/jackdelahunt/ai-edge/blob/08b7ffa743ecdd68a161fecdaa65105bccc2df36/test/e2e-tests/tests/pipelines_test.go#L23-L50) looked good as well. The new one won't fetch pipeline runs which are already completed every time the loop is waiting for a specific pipeline to finish. I think there is no nicer way of doing it.

Makefile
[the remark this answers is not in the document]
MarianMacik: My comment was that 2 secrets are created for image registry credentials but just one is really used - the one linked to the pipeline service account. In other words, the second sed is not needed AFAIK.

test/e2e-tests/tests/pipelines_test.go
```diff
@@ -0,0 +1,112 @@
...
+const (
+	AIEdgeE2EPipelineDirectoryRelativePath = "../../../pipelines/tekton/aiedge-e2e"
+	PipelineRunFileRelativePath            = AIEdgeE2EPipelineDirectoryRelativePath + "/aiedge-e2e.pipelinerun.yaml"
+)
+
+func CreateContext() context.Context {
+	return context.Background()
+}
+
+func WaitForAllPipelineRunsToComplete(ctx context.Context, clients *support.Clients) error {
```
MarianMacik, line 22: Actually here we can have a generic function waitFor() which would call this function and we don't need any sleeps here. We used something similar in the past with @sutaakar.
See here: https://github.com/kiegroup/kogito-operator/blob/main/test/pkg/framework/util.go#L83-L113
And more generic code here: https://gist.github.com/ngauthier/d6e6f80ce977bedca601

jackdelahunt: Updated with the generic solution and tested 👍

Makefile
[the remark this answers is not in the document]
jackdelahunt: The `IMAGE_REGISTRY_CREDENTIALS` file is only one secret and that is linked. I am not sure which secret you are referring to. I got the linking step from the README so maybe that is not needed actually? Or are you saying the username and password secrets are not required?

Makefile
```diff
@@ -34,7 +34,56 @@ endif
...
+	$(error S3_BUCKET is undefined)
+endif
+ifndef TARGET_IMAGE_REPO
+	$(error TARGET_IMAGE_REPO is undefined)
+endif
+	oc project test-namespace
+	(cd test/e2e-tests/tests && S3_BUCKET=${S3_BUCKET} TARGET_IMAGE_REPO=${TARGET_IMAGE_REPO} ${GO} test -timeout 30m)
+	oc delete project test-namespace
+
 test:
```
grdryn, line 87: When someone invokes `make test`, would they expect that it's just the shell tests that run now, or would they also expect the `go-test` to run?

Makefile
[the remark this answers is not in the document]
grdryn: > I think we should simplify this entire setup workflow to have the user provide their own unique registry & credentials files instead of using the Makefile to process them.

@LaVLaS I think the intention of this Make target is just for automated test setup: it's not intended to be used by users. Are you suggesting that it would be simpler to have the provisioning of these secrets not be part of `go-test-setup`?

Makefile
```diff
@@ -34,7 +34,56 @@ endif
...
+go-test: go-test-setup
+ifndef S3_BUCKET
+	$(error S3_BUCKET is undefined)
+endif
+ifndef TARGET_IMAGE_REPO
+	$(error TARGET_IMAGE_REPO is undefined)
+endif
+	oc project test-namespace
+	(cd test/e2e-tests/tests && S3_BUCKET=${S3_BUCKET} TARGET_IMAGE_REPO=${TARGET_IMAGE_REPO} ${GO} test -timeout 30m)
+	oc delete project test-namespace
```
grdryn, line 85: I'm not suggesting that this needs to be changed, just highlighting that it's a little asymmetrical (for lack of a better term) that this target doesn't create the namespace, but it does delete it.

Trying to understand @LaVLaS [comment](https://github.com/opendatahub-io/ai-edge/pull/188#discussion_r1478552193), should test setup and teardown should be decoupled from `go-test`? In essence, if I run `go-test-setup` once, should I be able to run `go-test` many times, and then eventually delete the namespace when I'm done?

If `go-test-setup` is required before _each_ invocation of `go-test`, then maybe there should only be `go-test`, and it should incorporate the existing recipe that's in `go-test-setup`?

Makefile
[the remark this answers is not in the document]
grdryn: I think what @MarianMacik is saying is that it's not clear why we need to create two secrets in these two lines in the `go-test-setup`:
```
@sed -e "s#{{ IMAGE_REGISTRY_USERNAME }}#${IMAGE_REGISTRY_USERNAME}#g" -e "s#{{ IMAGE_REGISTRY_PASSWORD }}#${IMAGE_REGISTRY_PASSWORD}#g" pipelines/tekton/aiedge-e2e/templates/credentials-image-registry.secret.yaml.template | oc create -f -
oc create -f ${IMAGE_REGISTRY_CREDENTIALS}
```

The first one is going to create a secret called `credentials-image-registry` based on the referenced template file, and the second is whatever is in the file referenced by `IMAGE_REGISTRY_CREDENTIALS`. Could/should both contain details to the same account? E.g. should `IMAGE_REGISTRY_USERNAME` and `IMAGE_REGISTRY_PASSWORD` be the quay robot account name and token? In which case, should there just be one secret? Or is there an expectation that they're accounts? Potentially different registries?

Makefile
jackdelahunt: Right now just shell tests. The target `go-test` is use for testing the pipeline locally with your own cluster. The target `test` is run from the ci which is where we do not want to run it

Makefile
jackdelahunt: This is what I was also confused about aswell when it was suggested to seperate them. Because the setup requires a namespace to be present as we are interacting with the cluster I feel like it is then a part of the tests aswell.

jackdelahunt: One thing maybe is that neither the test or test-setup targets delete or create the namespace and always use the current one in use

jackdelahunt:
I squashed most of the commits down to the actual testing and doc updates

Makefile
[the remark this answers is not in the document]
jackdelahunt: I think the biggest reason why these are not the same is because the `IMAGE_REGISTRY_CREDENTINALS` is gotten from quay itself as its own yaml file they provide, this will be unique for every user's account. The password and username template is fine as that is all we need to change for it, so it is not actually about the file.

But for the other one the file is what is actually significant, not sure if my thoughts made sense there

Makefile
grdryn: In that case, my suggestion for now, would be to not delete the namespace as part of this recipe (or change to a static `test-namespace` directory), and not have the `go-test-setup` one as a prerequisite. That would mean that someone could do something along the lines of the following:

```bash
make go-test-setup
make go-test
make go-test
...
```

jackdelahunt: Okay I get you, I think that shouldn't be that much of a change either

jackdelahunt: @LaVLaS was this what you were expecting, just want to make sure this is the setup we want

Makefile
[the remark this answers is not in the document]
grdryn: From https://github.com/opendatahub-io/ai-edge/tree/main/pipelines#setup-credentials:

> Image Registry - [credentials-image-registry.secret.yaml.template](https://github.com/opendatahub-io/ai-edge/blob/main/pipelines/tekton/aiedge-e2e/templates/credentials-image-registry.secret.yaml.template) to include the username and password with write access to the image repository. This is needed only in case you noted username and password of the robot account. In case you directly downloaded and applied a Kubernetes Secret, this file can be skipped.

So it sounds like _*either*_ you need to have a `.dockerconfigjson` secret file, similar to the one that you can get from a Quay robot account (there are a bunch of things you can get from a robot account, including username/password), and have that linked to the SA; _*or*_ create a secret from the template by substituting the template values and then linking that to the SA. Does that sound right?

The problem with this `go-test-setup` make target recipe, is that it's requiring both, and only linking the `IMAGE_REGISTRY_CREDENTIALS` one. If only one is required, for simplicity let's just use the simplest one (probably the robot file one).

WDYT?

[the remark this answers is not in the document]
jackdelahunt: > In case you directly downloaded and applied a Kubernetes Secret, this file can be skipped.

Agh okay I guess I missed this part when reading, now I understand the comments from before.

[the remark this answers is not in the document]
jackdelahunt: I think I would actually rather the username and password because then there is less manual work for running the tests, and having an extra file laying around is annoying. But maybe that isn't the only consideration.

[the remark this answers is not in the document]
grdryn: Sure, I'm easy with either :+1:

Makefile
jackdelahunt: Done 👍

test/e2e-tests/support/kustomize.go
```diff
@@ -0,0 +1,111 @@
...
+				if err != nil {
+					return err
+				}
+
+				_, err = clients.Kubernetes.CoreV1().ConfigMaps("test-namespace").Get(ctx, configMap.Namespace, metav1.GetOptions{})
+				if err == nil {
+					return nil
+				}
+
+				_, err = clients.Kubernetes.CoreV1().ConfigMaps("test-namespace").Create(ctx, &configMap, metav1.CreateOptions{})
```
grdryn, line 62: It looks like there's an implicit assumption here that the namespace is called `test-namespace`, but that is no longer being created by the Make recipe, right? Should/could this be changed to not rely on the namespace name being `test-namespace`?

test/e2e-tests/README.md
```diff
@@ -1,16 +1,33 @@
...
+
+The following enviroment varaibles are required to run the test setup and the tests themselves. If any of these are not set then the tests will not run. Read [here](../../pipelines/README.md#ai-edge-end-to-end-pipeline) for more context on how to set these up.
+
+- `AWS_SECRET_ACCESS_KEY` - Secret from AWS
+- `AWS_ACCESS_KEY_ID` - Access key from AWS
+- `S3_REGION` - Region of the bucket used to store the model
+- `S3_ENDPOINT` - Endpint of the bucket
+- `IMAGE_REGISTRY_USERNAME` - quay.io username
+- `IMAGE_REGISTRY_PASSWORD` - quay.io password
+- `IMAGE_REGISTRY_CREDENTIALS` - quay.io robot secret file
```
grdryn, line 20:
```suggestion
```

test/e2e-tests/README.md
```diff
@@ -1,16 +1,33 @@
...
 
-Place for testing and to verify the MLOps pipelines are working correctly. These tests use the `k8s.io/client-go` package to interact with the cluster. Using `oc login` to log into the cluster as normal should mean no setup is needed to run the tests.
+Place for testing and to verify the MLOps pipelines are working correctly. These tests use the `k8s.io/client-go` package to interact with the cluster. Using `oc login` to log into the cluster as normal should mean little setup is needed to run the tests.
 
 A local install of the Go compiler is needed to run the tests. Go version `1.21` is required to run the tests.
 
+## Setup
+- Log into the target cluster using `oc login`. This will update your default `kubeconfig` for the tests to use
+- Create a S3 bucket with the models in the root directory
+- Create a quay.io robot
```
grdryn, line 10:
```suggestion
```

This one isn't a hard requirement. While it will be what is usually done, someone could pass a user's username/password, or use a different registry that doesn't have a concept of robot accounts.

test/e2e-tests/README.md
```diff
@@ -1,16 +1,33 @@
...
+
+- `AWS_SECRET_ACCESS_KEY` - Secret from AWS
+- `AWS_ACCESS_KEY_ID` - Access key from AWS
+- `S3_REGION` - Region of the bucket used to store the model
+- `S3_ENDPOINT` - Endpint of the bucket
+- `IMAGE_REGISTRY_USERNAME` - quay.io username
+- `IMAGE_REGISTRY_PASSWORD` - quay.io password
+- `IMAGE_REGISTRY_CREDENTIALS` - quay.io robot secret file
+- `S3_BUCKET` - Name of S3 bucket that has the model
+- `TARGET_IMAGE_REPO` - Image repository that built model will be pushed to
```
grdryn, line 22:
```suggestion
- `TARGET_IMAGE_REPO` - Image registry namespace that the built model will be pushed to a repository in (the repository name will be the same as the model name).
```

The pipeline parameter is confusingly named (and described). My suggestion here is a more accurate description. That pipeline parameter will be replaced as part of [RHOAIENG-2758](https://issues.redhat.com/browse/RHOAIENG-2758), so I wouldn't bother changing anything else around it in this PR, apart from this description here since this is new.

grdryn (commented):
Some final comments here. Once these are addressed (esp. the hardcoded namespace one), this will be all good for me. Thanks!

Makefile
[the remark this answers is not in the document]
LaVLaS: Adding a comment here based on offline conversations, we will need to refactor this in follow-up PRs to separate the functionality into
1. Creation and application of the user objects that are based on the templates we provide (S3 credentials, image registry, test-data,...).  This should probably be a separate script that could be used for users and CI to ensure everything is processed and applied correctly to the server
2. Setup that is required specifically for the go-tests.

Right now we have divergent paths for the [shell tests](test/shell-pipeline-tests) and the functional tests proposed in this PR so we need to get to a state where we are utilizing the same workflow for e2e and functional tests

test/e2e-tests/README.md
jackdelahunt: I renamed this to `TARGET_IMAGE_NAMESPACE`

test/e2e-tests/support/kustomize.go
jackdelahunt: I tried to find a way to get the current active namespace that is set in the kubeconfig from oc but couldn't figure it out. I created a namespace parameter for the tests instead

test/e2e-tests/support/kustomize.go
```diff
@@ -0,0 +1,111 @@
...
+		switch kind {
+		case "ConfigMap":
+			{
+				var configMap corev1.ConfigMap
+				err := ResourceToType(rsc, &configMap)
+				if err != nil {
+					return err
+				}
+
+				_, err = clients.Kubernetes.CoreV1().ConfigMaps(namespace).Get(ctx, configMap.Namespace, metav1.GetOptions{})
```
grdryn, line 57: I didn't notice it until the changes to parameterise the namespace, but it seems odd that the namespace is needed to create the `Clients` instance, and then is still also needed for functions that are based on that. My feeling is that our Clients abstraction could be improved in one direction or the other.

grdryn (approved):
I've added one final nit in an inline comment, but it's something that can be improved in a follow-on PR. It's not enough to hold up this PR any longer.

Thanks @jackdelahunt! :slightly_smiling_face:

Makefile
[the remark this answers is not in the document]
MarianMacik: @LaVLaS I think the approach is the same, we use the same templates for both go and shell tests. Just for Openshift CI the credentials need to be obtained from secrets.

PR #191, opened 2023-12-13
RHOAIENG-905 - Create a PR check for testing README file (without gitops)

MarianMacik:
…ops)

**JIRA:** [RHOAIENG-905](https://issues.redhat.com/browse/RHOAIENG-905)

## Description
Adds 2 scripts:
* common.sh which acts as a library
* pipeline-test.sh currently testing just the build-container-image-pipeline and test-mlflow-image-pipeline with housing (TensorFlow) version of the pipeline run

## How Has This Been Tested?
Locally using a separate cluster.

## Merge criteria:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,4 +32,6 @@ GO=go
 GOFLAGS=""
 
 test:
+	@(./test/shell-pipeline-tests/openvino-bike-rentals/pipelines-test-openvino-bike-rentals.sh)
+	@(./test/shell-pipeline-tests/tensorflow-housing/pipelines-test-tensorflow-housing.sh)
 	@(cd test/e2e-tests/tests && ${GO} test)
--- a/test/shell-pipeline-tests/common.sh
+++ b/test/shell-pipeline-tests/common.sh
@@ -28,7 +28,7 @@ function waitForOpResult() {
 function saveArtifacts() {
     ## Backup all Pipeline Runs, Task Runs
     local PIPELINE_RUN_NAME=$1
-    local LOGS_DIR="${ARTIFACT_DIR}/$PIPELINE_RUN_NAME"
+    local LOGS_DIR="${ARTIFACT_DIR}/$PIPELINE_RUN_NAME" # ARTIFACT_DIR is an env var on OpenShift-CI
     mkdir -p "$LOGS_DIR"
     echo "Archiving YAML definitions and logs for '$PIPELINE_RUN_NAME' to '$LOGS_DIR'"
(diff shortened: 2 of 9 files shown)


LaVLaS:
@MarianMacik Can you add a link to the Jira issue in the description for anyone who want to check for additional context in the jira issue?

Makefile
```diff
@@ -32,4 +32,6 @@ GO=go
 GOFLAGS=""
 
 test:
```
winklerm, line 34: As we discussed earlier, this target could be split into a target for the Go tests and a target for the shell tests, so they could be both triggered independently, if needed. The `test` target could call both of them.

This can be done later, just mentioning it here so it does not get forgotten.

test/shell-pipeline-tests/common.sh
```diff
@@ -0,0 +1,59 @@
...
+    oc get task -o yaml > "${LOGS_DIR}"/tasks.txt
+    oc get taskrun -l "tekton.dev/pipelineRun=$PIPELINE_RUN_NAME" -o yaml > "${LOGS_DIR}"/taskRuns.txt
+    oc logs -l "tekton.dev/pipelineRun=$PIPELINE_RUN_NAME" --all-containers --prefix --tail=-1 > "${LOGS_DIR}"/logs.txt
+}
+
+function createS3Secret() {
+    local AWS_SECRET_PATH_TEMPLATE=$1
+    local AWS_SECRET_PATH=$2
+
+    local AI_EDGE_AWS_VAULT_OPENSHIFT_CI_SECRET_PATH
```
LaVLaS, line 45: Based on this naming convention, does this require that running this locally requires setting up the secrets in the format that is compatible with openshift-ci?

Makefile
MarianMacik: I agree, this will be addressed in the next PRs.

test/shell-pipeline-tests/common.sh
MarianMacik: @LaVLaS yes, you are right. I added the readme to guide what needs to be done. It is just 3 variables which need to be set. In the future, we may even provide some defaults in the Makefile.

winklerm:
Looks good to me, not approving now in case @LaVLaS has additional comments.

Makefile
LaVLaS: Since this will be executing two separate frameworks, should we go head a make a `test-go` target to separate them now

LaVLaS:
/lgtm

Changes look reasonable to me.

PR #192, opened 2023-12-20
RHOAIENG-203: Use ArgoCD pull controller integration for GitOps

grdryn:
JIRA: https://issues.redhat.com/browse/RHOAIENG-203

## Description
This change moves the GitOps machinery from the ACM AppSubscription model that was used here before, to ACM's integration with the ArgoCD/OpenShift Pipelines pull controller. No changes are needed in the referenced gitops repositories.

## How Has This Been Tested?

I've added some kustomize tests to validate that all modifications occur as expected. These can be checked by witnessing that the following command exits with a return code of 0:
```
kustomize build acm/registration/near-edge/test/
```

I have also verified that both examples deploy as expected, and can be deleted as expected also. This can be done using the following command:
```
oc apply -k acm/registration/
```
and then once satisfied, deletion can be achieved using the following:
```
oc delete -k acm/registration/
```

I also tested that ACM section in DEVELOPMENT.md, and associated changes to Makefile and YAML files, works as expected.

## Merge criteria:

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -58,42 +58,29 @@ We use a mix of AWS and GCP to host these clusters for diversity. Typically you 
 
 * Create a new testing branch on your fork of the `opendatahub-io/ai-edge` repo, preferably named `<your-kerberos-name>-acm-dev`.
 * Ensure that you change `namespace` in either `test/gitops/bike-rental-app/kustomization.yaml` or `test/gitops/tensorflow-housing-app/kustomization.yaml`, or both, as appropriate.
-* Substitute the values of the `GIT_REPO_URL`, `GIT_BRANCH`, and `TEST_NAMESPACE` variables in this `make` command appropriately:
+* Substitute the values of the `GIT_REPO_URL`, `GIT_BRANCH`, `CUSTOM_PREFIX`, and `CUSTOM_APP_NAMESPACE` variables in this `make` command appropriately:
 NOTE: Escape the `:` in the `https://` protocol part of the `GIT_REPO_URL` value
 ```bash
 make -s -e GIT_REPO_URL="https\://github.com/opendatahub-io/ai-edge" \
-     -e GIT_BRANCH=my-git-branch \
-     -e TEST_NAMESPACE=my-test-namespace \
+     GIT_BRANCH=my-git-branch \
+     CUSTOM_PREFIX=custom-prefix- \
+     CUSTOM_APP_NAMESPACE=my-test-namespace \
      test-acm-bike-rental-app-generate # or test-acm-tensorflow-housing-generate
 ```
 * You can also do a dry-run apply of these manifests, by piping the output to `oc apply -f - --dry-run=client`, like this:
 
 ```bash
 make -s -e GIT_REPO_URL="https\://github.com/opendatahub-io/ai-edge" \
-     -e GIT_BRANCH=my-git-branch \
-     -e TEST_NAMESPACE=my-test-namespace \
+     GIT_BRANCH=my-git-branch \
+     CUSTOM_PREFIX=custom-prefix- \
+     CUSTOM_APP_NAMESPACE=my-test-namespace \
      test-acm-bike-rental-app-generate # or test-acm-tensorflow-housing-generate | oc apply -f - --dry-run=client
 ```
 * If everything looks correct, run the make target again and apply the manifests to the hub cluster without the dry-run option
 ```bash
 make -s -e GIT_REPO_URL="https\://github.com/opendatahub-io/ai-edge" \
-     -e GIT_BRANCH=my-git-branch \
-     -e TEST_NAMESPACE=my-test-namespace \
+     GIT_BRANCH=my-git-branch \
+     CUSTOM_PREFIX=custom-prefix- \
+     CUSTOM_APP_NAMESPACE=my-test-namespace \
      test-acm-bike-rental-app-generate  | oc apply -f -
 ```
-* Add your dev namespace to the near-edge Cluster Set's `Namespace Bindings`.
-  * In the OpenShift Console, navigate to `All Clusters (dropdown in top left) -> Infrastructure -> Clusters -> Cluster sets -> <cluster-set-name> -> Namespace Bindings`.
-
-      OR
-
-  * ```bash
-      cat << EOF | oc apply -f -
-      apiVersion: cluster.open-cluster-management.io/v1beta2
-      kind: ManagedClusterSetBinding
-      metadata:
-         name: ___CLUSTER_SET_NAME___
-         namespace: ___DEV_NAMESPACE___
-      spec:
-         clusterSet: ___CLUSTER_SET_NAME___
-      EOF
-      ```
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ based on the Open Cluster Management project.
 | Red Hat OpenShift Pipelines                             | 1.11 or higher |
 | Quay Registry account                                   | [quay.io](https://quay.io/) |
 | Advanced Cluster Management for Kubernetes              | 2.8            |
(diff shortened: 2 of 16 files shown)


DEVELOPMENT.md
```diff
@@ -58,42 +58,29 @@ We use a mix of AWS and GCP to host these clusters for diversity. Typically you
 
 * Create a new testing branch on your fork of the `opendatahub-io/ai-edge` repo, preferably named `<your-kerberos-name>-acm-dev`.
 * Ensure that you change `namespace` in either `test/gitops/bike-rental-app/kustomization.yaml` or `test/gitops/tensorflow-housing-app/kustomization.yaml`, or both, as appropriate.
-* Substitute the values of the `GIT_REPO_URL`, `GIT_BRANCH`, and `TEST_NAMESPACE` variables in this `make` command appropriately:
+* Substitute the values of the `GIT_REPO_URL`, `GIT_BRANCH`, `CUSTOM_PREFIX`, and `CUSTOM_APP_NAMESPACE` variables in this `make` command appropriately:
```
grdryn, line 61: I changed `TEST_NAMESPACE` to `CUSTOM_APP_NAMESPACE` here because previously we used a shared namespace name for resources on both the hub and managed clusters, whereas with this change, the resources on the hub are applied to the `openshift-gitops` namespace.

Given that everything on the hub goes to the `openshift-gitops`, `CUSTOM_PREFIX` is needed to prevent clobbering resources from other instances of the app that you may have deployed. All of this might be less necessary now than it was when we used a shared dev cluster, but I can imagine a scenario where you've got an instance deployed when you're working on something, but then you may want to deploy an instance to test out changes in a PR from a colleague.

README.md
```diff
@@ -121,19 +108,66 @@ and the applications in those remote git repositories will be deployed.
 By default, [github.com/opendatahub-io/ai-edge](https://github.com/opendatahub-io/ai-edge) is configured;
 edit the URLs to match your repositories.
 
-Then run as a regular (non-admin) user
+Then, as an admin user, run:
```
grdryn, line 111: This is an interesting part: it will at least have to be a user with permissions on ApplicationSets in the `openshift-gitops` namespace. This may have UX/UI implications when we go to integrate it more tightly with Open Data Hub.

acm/registration/near-edge/base/near-edge.yaml
```diff
@@ -1,39 +1,36 @@
 ---
-apiVersion: app.k8s.io/v1beta1
-kind: Application
+apiVersion: policy.open-cluster-management.io/v1
+kind: Policy
 metadata:
-  name: application
-  namespace: near-edge-acm-template
+  name: namespace-policy
+  namespace: openshift-gitops
```
grdryn, line 6: This per-application `Policy` resource doesn't really need to be in `openshift-gitops`, and maybe this isn't the best place for it on the hub. Open to suggestions around that, whether it's okay, or whether it should go to some other namespace. I mostly chose here for now, as that's where the closely-related ApplicationSet resource needs to be.

acm/registration/near-edge/kustomization.yaml
```diff
@@ -2,5 +2,8 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
 resources:
+- openshift-gitops-managedclustersetbinding.yaml
+- openshift-gitops-placement.yaml
+- openshift-gitops-gitopscluster.yaml
```
grdryn, line 7: These 3 resources are not really _per application_: they're more just needed as prerequisites needed before any application.

acm/registration/near-edge/openshift-gitops-placement.yaml
```diff
@@ -0,0 +1,17 @@
+---
+apiVersion: cluster.open-cluster-management.io/v1beta1
+kind: Placement
+metadata:
+  name: near-edge-clusters
+  namespace: openshift-gitops
+spec:
+  clusterSets:
+    - poc-near-edge
```
grdryn, line 9: This Placement is for the `GitOpsCluster` and doesn't necessarily need to be limited to the same cluster set. It could contain the `global` cluster set, so that different GitOpsClusters aren't needed for different configurations of target clusters.

If this was matched to the `global` cluster set, then we'd need a `ManagedClusterSetBinding` for the `global` cluster set in `openshift-gitops`, and then additional `ManagedClusterSetBindings` for any other cluster sets that are referenced in other placements that are created and referenced in `ApplicationSet` or `Policy` resources.

grdryn (commented):
Self review, just to highlight some things.

README.md
LaVLaS: Which level of Admin user is required, cluster or project?

grdryn: I've updated it to try to make it clearer that what's needed is a user with permissions on certain resources in the openshift-gitops namespace. Please let me know if the new change doesn't clarify things for you, thanks :)

test/acm/bike-rental-app/kustomization.yaml
```diff
@@ -4,22 +4,62 @@ kind: Kustomization
...
-namespace: my-test-namespace
+namePrefix: custom-prefix-
 
 patches:
 - patch: |-
     - op: replace
-      path: /spec/channel
-      value: my-test-namespace/bike-rental-app-channel
+      path: /spec/template/metadata/name
+      value: '{{name}}-grdryn-bike-rental-app'
```
LaVLaS, line 13: I assume your username was not intended to be included?

test/acm/tensorflow-housing-app/kustomization.yaml
```diff
@@ -4,22 +4,62 @@ kind: Kustomization
...
-namespace: my-test-namespace
+namePrefix: custom-prefix-
 
 patches:
 - patch: |-
     - op: replace
-      path: /spec/channel
-      value: my-test-namespace/tensorflow-housing-app-channel
+      path: /spec/template/metadata/name
+      value: '{{name}}-grdryn-tensorflow-housing-app'
```
LaVLaS, line 13: I assume your username was not intended to be included?

test/acm/bike-rental-app/kustomization.yaml
grdryn: Correct, good catch, thanks! I've fixed it now.

Related to this, I'm not sure how useful these test/dev kustomizations are now, given that we don't all share a single cluster. Maybe once this PR is merged, I'll propose removing them in a separate PR, to see if there are any objections or if there's anybody still using them.

LaVLaS:
@grdryn Is the use of the `{{name}}` intended to be replaced by `kustomize` or some other command?

When I run `kustomize build acm/registration/near-edge/test/`, the `ApplicationSet` as the following for the annotation and labels

```
      annotations:                                                                                    
        apps.open-cluster-management.io/ocm-managed-cluster: '{{name}}'
        apps.open-cluster-management.io/ocm-managed-cluster-app-namespace: openshift-gitops
        argocd.argoproj.io/skip-reconcile: "true"
      labels:                                  
        apps.open-cluster-management.io/pull-to-ocm-managed-cluster: "true"
      name: '{{name}}-tensorflow-housing-app'   
```

grdryn:
> Is the use of the {{name}} intended to be replaced by kustomize or some other command?

@LaVLaS That's intended to be replaced at run time rather than build time: the ApplicationSet controller will create an Applications based on the template specified in the ApplicationSet, one per cluster. It will generate a list of cluster details based on the config specified in the "generators" section of the ApplicationSet spec, and replace those template parameters with values from the cluster details in the list.

Here's some more info:
https://github.com/argoproj/argo-cd/blob/3224102664b747e88e7330e59b3b44bc9c28cd66/docs/operator-manual/applicationset/index.md#parameter-substitution-into-templates

and:
https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.9/html/gitops/gitops-overview#crd-pull-model

PR #194, opened 2024-01-12
Add list of servers for supported model formats

grdryn:
JIRA: https://issues.redhat.com/browse/RHOAIENG-290

## Description
The first commit here just gets `yq` to reformat the YAML files to its preferred style, so that the actual changes are clearer in the second commit. A YAML diff tool, such as https://yamldiff.com/, can be used to verify that the before and after of this first commit have the same YAML meaning. Adding the `w=1` URL parameter to the files view here on GitHub will also minimize most of the formatting noise).

The second commit introduces a new file that has a list of mappings of model file formats to model servers. This file is just the source for the annotation value in the Pipeline and PipelineRun resources in the YAML files. When it's changed, the annotations can be updated using the following commands:

```sh
for i in aiedge-e2e.pipeline.yaml aiedge-e2e.pipelinerun.yaml ; do
    yq -i e ".metadata.annotations[\"opendatahub.io/supported-model-formats\"] = $(cat supported-model-formats.json | jq '@json')" $i
done
```

The list can be queried in a manner similar to the following:
```sh
oc get pipeline aiedge-e2e -o json \
   | jq '.metadata.annotations["opendatahub.io/supported-model-formats"] | fromjson | .[] | select(."model-format" == "onnx")'
```
## How Has This Been Tested?

I successfully deployed this to a cluster, and successfully queried the list from the annotation using the command listed in the description above.

## Merge criteria:

### DIFF
--- a/pipelines/tekton/aiedge-e2e/aiedge-e2e.pipelinerun.yaml
+++ b/pipelines/tekton/aiedge-e2e/aiedge-e2e.pipelinerun.yaml
@@ -6,52 +6,52 @@ metadata:
   generateName: aiedge-e2e-
 spec:
   params:
-  - name: model-name
-    value: bike-rentals-auto-ml
-  - name: model-version
-    value: "1"
-  - name: s3-bucket-name
-    value: rhoai-edge-models
-  - name: gitServer
-    value: https://github.com
-  - name: gitOrgName
-    value: opendatahub-io
-  - name: gitRepoName
-    value: ai-edge
-  - name: containerfileRelativePath
-    value: pipelines/containerfiles/Containerfile.seldonio.mlserver.mlflow
-  - name: fetch-model
-    value: "s3"
-  - name: git-model-repo
-    value: https://github.com/opendatahub-io/ai-edge.git
-  - name: modelRelativePath
-    value: pipelines/models/
-  - name: git-revision
-    value: "main"
-  - name: test-endpoint
-    value: "invocations"
-  - name: target-imagerepo
-    value: rhoai-edge
+    - name: model-name
+      value: bike-rentals-auto-ml
+    - name: model-version
+      value: "1"
+    - name: s3-bucket-name
+      value: rhoai-edge-models
+    - name: gitServer
+      value: https://github.com
+    - name: gitOrgName
+      value: opendatahub-io
+    - name: gitRepoName
+      value: ai-edge
+    - name: containerfileRelativePath
+      value: pipelines/containerfiles/Containerfile.seldonio.mlserver.mlflow
+    - name: fetch-model
+      value: "s3"
+    - name: git-model-repo
+      value: https://github.com/opendatahub-io/ai-edge.git
+    - name: modelRelativePath
+      value: pipelines/models/
+    - name: git-revision
+      value: "main"
+    - name: test-endpoint
+      value: "invocations"
+    - name: target-imagerepo
+      value: rhoai-edge
   pipelineRef:
     name: aiedge-e2e
   serviceAccountName: pipeline
   timeout: 1h0m0s
   workspaces:
-  - name: buildah-cache
-    # Have the tekton controller allocate a PVC for each pipeline run that persists for the life of each PipelineRun object.
-    # NOTE: This PVC will be deleted by the Tekton controller when the PipelineRun is deleted
-    volumeClaimTemplate:
-      spec:
-        accessModes:
-          - ReadWriteOnce
-        resources:
-          requests:
-            storage: 1Gi
-  - name: s3-secret
-    secret:
+    - name: buildah-cache
+      # Have the tekton controller allocate a PVC for each pipeline run that persists for the life of each PipelineRun object.
+      # NOTE: This PVC will be deleted by the Tekton controller when the PipelineRun is deleted
+      volumeClaimTemplate:
+        spec:
+          accessModes:
+            - ReadWriteOnce
+          resources:
+            requests:
+              storage: 1Gi
+    - name: s3-secret
+      secret:
         secretName: credentials-s3
-  - emptyDir: {}
-    name: workspace
-  - configMap:
-      name: bike-rentals-test-data
-    name: test-data
+    - emptyDir: {}
+      name: workspace
+    - configMap:
+        name: bike-rentals-test-data
+      name: test-data
--- a/pipelines/tekton/aiedge-e2e/supported-model-formats.json
+++ b/pipelines/tekton/aiedge-e2e/supported-model-formats.json
@@ -1,0 +1,30 @@
+[
+    {
+        "model-format": "tensorflow",
+        "model-server": "ovms"
+    },
+    {
+        "model-format": "tensorflow-lite",
+        "model-server": "ovms"
+    },
+    {
+        "model-format": "onnx",
+        "model-server": "ovms"
+    },
+    {
+        "model-format": "pytorch",
+        "model-server": "ovms"
+    },
+    {
+        "model-format": "openvino-ir",
+        "model-server": "ovms"
+    },
+    {
+        "model-format": "scikit-learn",
+        "runtime": "seldon"
+    },
+    {
+        "model-format": "mlflow",
+        "runtime": "seldon"
+    }
+]
--- a/pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
+++ b/pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
@@ -4,377 +4,374 @@ metadata:
   name: aiedge-e2e
 spec:
   params:
-  - name: model-name
-    type: string
-    description: Name of the directory where the model files are stored
-  - default: "1"
-    name: model-version
-    type: string
-    description: Version of the model
-  - name: s3-bucket-name
-    type: string
-    description: "[S3 ONLY] S3 bucket name where the model is stored"
-    default: ""
-  - name: gitServer
-    type: string
-    description: Root domain of the Git server
-  - name: gitOrgName
-    type: string
-    description: Organization or Group name where the repo is located
-  - name: gitRepoName
-    type: string
-    description: Git repository name
-  - name: containerfileRelativePath
-    type: string
-    description: Path to the Containerfile in the git repository for model server imagebuild
-  - name: modelRelativePath
-    type: string
-    description: "Location of the model within the context of the source location.  Leave blank if the model files are at the root of the source location."
-  - name: fetch-model
-    type: string
-    description: "Fetch the model from Git or S3. Valid values in [s3, git]"
-  - name: git-model-repo
-    type: string
-    description: The git repo url where the model files are stored
-    default: ""
-  - name: git-revision
-    type: string
-    description: The Git ref to use when cloning the git repo with the saved model
-    default: "main"
-  - name: test-endpoint
-    type: string
-    description: The inferencing endpoint for the model to use for testing
-  - default: $(context.pipelineRun.namespace)
-    name: target-namespace
-    type: string
-  - name: target-imagerepo
-    type: string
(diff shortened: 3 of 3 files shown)


pipelines/tekton/aiedge-e2e/supported-model-formats.json
```diff
@@ -0,0 +1,30 @@
+[
```
grdryn, line 1: Questions for reviewers:
- Does this list make sense?
- Do the keys make sense, or is there something else you'd suggest?
- Do the values make sense, or are there alternative values that you'd suggest?
- Are there other elements that should be added to the list, from the supported model types by either model server? I haven't added all of the types that [Seldon lists support for](https://docs.seldon.io/projects/seldon-core/en/latest/nav/config/servers.html), should I? Would the different supported Seldon model types require different base images, or will there just be one? The [Seldon docs](https://docs.seldon.io/projects/seldon-core/en/latest/nav/config/servers.html) appear to list different "Servers" for the different types

pipelines/tekton/aiedge-e2e/supported-model-formats.json
```diff
@@ -0,0 +1,30 @@
...
+    {
+        "model-format": "pytorch",
+        "model-server": "ovms"
+    },
+    {
+        "model-format": "openvino-ir",
+        "model-server": "ovms"
+    },
+    {
+        "model-format": "scikit-learn",
```
grdryn, line 23: Should this be `joblib`? or `scikit-learn-joblib` or something? https://docs.seldon.io/projects/seldon-core/en/latest/servers/sklearn.html

Also, if I understand things correctly, then scikit-learn models can also be stored in mlflow format, so if we listed them as separate options, which should a user choose?

grdryn (commented):
Self review to add questions inline for others.

pipelines/tekton/aiedge-e2e/supported-model-formats.json
```diff
@@ -0,0 +1,30 @@
+[
+    {
+        "model-format": "tensorflow",
+        "model-server": "ovms"
```
grdryn, line 4: Seldon also seems to support Tensorflow. Would there ever be a case that a user would want to explicitly use that model server for Tensorflow (I guess there would be a few)? If so, then maybe we shouldn't try to force them to use the OpenVino one, right? Maybe we should just expose the model servers that we support, and make the user choose _that_, rather than telling us about the model and then having us choose?

israel-hdez: I haven't heard anything from the field, but I think it is going to be very likely that users may want to use a specific one. My guess is that the choose may be because of the varying features each model server.

E.g. if a given model would run much faster on a GPU (vs a CPU), OpenVino would be the natural choice if the user infrastructure has Intel GPUs, while Seldon would be the choice for nVidia GPUs. Then, if the infra is mixed, the user may want to build two images containing the same model but using different model servers. This is just an example which may be not very valid for edge-use cases, assuming GPUs would be rare, but I think it works to show that we may need flexibility here.

israel-hdez: > Maybe we should just expose the model servers that we support, and make the user choose _that_, rather than telling us about the model and then having us choose?

BTW, this is how the odh-dashboard currently works; i.e. you need to choose which model server you want to use when serving a model. For edge, it would be a pipeline rather than directly a model server, but the UX would be quite similar, I think, if we go this route --- So:

* Non-edge case: User picks a model server to serve the desired model
* Edge use case: User picks the pipeline (which inherently means to pick a server) to package the desired model.

I would find the UX so similar up to the point that some users may not realize it is technically different.

pipelines/tekton/aiedge-e2e/supported-model-formats.json
israel-hdez: > Should this be `joblib`? or `scikit-learn-joblib` or something? https://docs.seldon.io/projects/seldon-core/en/latest/servers/sklearn.html
> 

You may want to check this sources of KServe project: https://github.com/kserve/kserve/tree/8357567e8644840613bb31ec37f5813879557c41/config/runtimes. Look under `supportedModelFormats` (e.g. [sklearn](https://github.com/kserve/kserve/blob/8357567e8644840613bb31ec37f5813879557c41/config/runtimes/kserve-sklearnserver.yaml#L9-L13))

We've have adopted KServe for non-edge cases. In ODH fork, we don't officially support any of those runtimes, but perhaps we may want naming consistency, since there are plans to support only a few of those servers. If we use same naming, this should be `sklearn`.

Notice that in KServe they use two fields: format name, and format version. The version distinction seems important for some formats (e.g. [tensorflow](https://www.tensorflow.org/hub/model_formats) has TF1, TF2, TF.js, TF-Lite). I don't think we need to use two fields here, but we need to make the distinction (e.g. you already used `model-format: tensorflow-lite` in an entry).

> Also, if I understand things correctly, then scikit-learn models can also be stored in mlflow format, so if we listed them as separate options, which should a user choose? This stuff confuses me.

I think this is similar to the other question and answer would be the same: user would pick based on the needed/preferred model server features.

pipelines/tekton/aiedge-e2e/supported-model-formats.json
israel-hdez: About the first 3 bullets, perhaps check my other replies to see if that clarifies something.

About the last one, I', OK if we add only what we need for now. About Seldon's base images... well... I don't know if this is Seldon-specific or if it is similar for other model servers, but yes, we may need to be able to use different base images. Indeed, in the sample models that Davide added, different Seldon images are being used. That said, I'd see the different Seldon images as different model servers.

pipelines/tekton/aiedge-e2e/aiedge-e2e.pipelinerun.yaml
```diff
@@ -4,54 +4,56 @@ metadata:
   labels:
     tekton.dev/pipeline: aiedge-e2e
   generateName: aiedge-e2e-
+  annotations:
```
LaVLaS, line 7: You don't need to define the duplicate annotations in the PipelineRuns.

Labels and annotations will [propagate](https://tekton.dev/docs/pipelines/labels/) from `Pipeline -> PipelineRun -> TaskRun`  with any labels on annotations defined in the `*Run` object taking precedence.

pipelines/tekton/aiedge-e2e/supported-model-formats.json
LaVLaS: Given that the PipelineRun is a singular choice of `model-server -> model-format`, we may need to change the structure to the model server as the primary key with a list of supported model formats 
```
{
   "<Model ServerX>": [ <Model Format>, <Model Format>, ...],
   "<Model ServerY>": [ <Model Format>, <Model Format>, ...],
   ...
   "<Model ServerZ>": [ <Model Format>, <Model Format>, ...]
}
```

@israel-hdez Do we support Seldon in ODH?  I know it was used in the PoC original since we inherited the bike-rental and tensorflow-housing models but I keep forgetting if it is one of the fully supported model serving runtimes in ODH

LaVLaS:
> The second commit introduces a new file that has a list of mappings of model file formats to model servers.

IMO, we don't need to maintain a separate file to host the mappings.  In another comment I pointed out that explicit `PipelineRun` annotations are not needed if they are duplicates from the referenced `Pipeline` object.  

See Tekton Propagation of [Labels and Annotations](https://tekton.dev/docs/pipelines/labels/)

pipelines/tekton/aiedge-e2e/supported-model-formats.json
grdryn: @israel-hdez For the edge case I think it will be a little different than what you describe, since (at least currently) there's just one pipeline, and the user needs to choose by entering an appropriate parameter value. Based on the context you've provided here, maybe it would be enough to just have a list instead of the map that I've proposed here so far. E.g. something like:

```js
["openvino", "seldon-sklearn", "seldon-mlflow", "seldon-tensorflow", "seldon-triton", "seldon-xgboost", "seldon-tempo"]
```

This way the UI would be able to display meaningful values, but also be able to select the correct base image / Containerfile for the pipeline run.

In the above example, I assume that there would only be one OpenVino base image for all of the model formats that it supports, and no additional configuration is needed for each model format?

pipelines/tekton/aiedge-e2e/supported-model-formats.json
grdryn: > In ODH fork, we don't officially support any of those runtimes, but perhaps we may want naming consistency, since there are plans to support only a few of those servers.

I don't fully understand this bit: I guess you're saying that using KServe in ODH & downstream, we don't support all of the model formats that KServe supports? How does a user determine whether a particular format is supported in ODH or not? What about for OpenVino & Seldon, are there particular model formats that we should or shouldn't support for these for these edge use cases?

pipelines/tekton/aiedge-e2e/supported-model-formats.json
grdryn: > Given that the PipelineRun is a singular choice of `model-server -> model-format`, we may need to change the structure to the model server as the primary key with a list of supported model formats
> 
> ```
> {
>    "<Model ServerX>": [ <Model Format>, <Model Format>, ...],
>    "<Model ServerY>": [ <Model Format>, <Model Format>, ...],
>    ...
>    "<Model ServerZ>": [ <Model Format>, <Model Format>, ...]
> }
> ```

I guess it depends on what it's needed for: I think there's potentially a case for the other direction too. I wonder if maybe we should delay doing this at all for now, until we get back to thinking about UI again at a later stage?

pipelines/tekton/aiedge-e2e/aiedge-e2e.pipelinerun.yaml
grdryn: Fixed, thanks.

pipelines/tekton/aiedge-e2e/supported-model-formats.json
israel-hdez: @LaVLaS We don't support Seldon at the moment. Only OpenVINO and, now, we also have Caikit/TGIS.

pipelines/tekton/aiedge-e2e/supported-model-formats.json
israel-hdez: AFAIK, in KServe we don't support specific formats. What we support are the Model Servers (aka Serving Runtimes). To get support, users would need to save their models in a format that the supported model servers can load.

In ODH, at the moment, the only two Model Servers we formally support are OpenVINO and Caikit/TGIS. 

That said, KServe/ModelMesh are agnostic to formats and servers. Because of the limited number of servers we currently support (and, in transition, the small set of formats), I've heard some users registering their preferred servers into KServe/ModelMesh to use them. We won't support their model server, but we will help with the rest of the stack.

Not sure how much this situation will change, given the edge project seems to need the other model servers.  Minimally, I see that need of picking Seldon.

MarianMacik:
@grdryn thanks, the job's green now.

pipelines/tekton/aiedge-e2e/supported-model-formats.json
grdryn: Thanks for the info @israel-hdez.

> In ODH, at the moment, the only two Model Servers we formally support are OpenVINO and Caikit/TGIS.

I'm not familiar with Caikit/TGIS: is that something that would be appropriate for edge, and if so, is there some appropriate example model that we could have here?

> Not sure how much this situation will change, given the edge project seems to need the other model servers. Minimally, I see that need of picking Seldon.

That's something that we should bottom out: are you saying the edge project seems to need Seldon just because we've got examples that use it, or do you think it's a model server that users will explicitly want to use in edge environments?

grdryn:
We're going to defer this for now.

PR #198, opened 2024-01-26
RHOAIENG-1373 - Remove the individual build and test pipelines

MarianMacik:
**JIRA:** [RHOAIENG-1373](https://issues.redhat.com/browse/RHOAIENG-1373)

## Description
* Build & Test pipelines have been moved under the `/pipelines/tekton/obsolete/` directory with a separate README file.
* aiedge-e2e README parts were moved to the parent README under the `/pipelines/tekton/` directory and the original README was removed.
* Minor improvements to the secret naming and various other documentation fixes.

## How Has This Been Tested?
It's mainly documentation update, so no need for any automated testing.

## Merge criteria:


README.md
```diff
@@ -212,7 +212,7 @@ You can set the `SIZE` and `PVC` values aswell
 make MODEL_PATH="PATH_TO_A_FILE" NAME=my-model SIZE=1G PVC=my-new-PVC create
 ```
 
-In `pipelines/tekton/build-container-image-pipeline` you can set the
+In `pipelines/tekton/obsolete/build-container-image-pipeline` you can set the
 `fetch-model` to `pvc` for the pipeline to copy the file from the
 `model-workspace` which can be set to the PVC created in the last step
```
MarianMacik, line 217: @LaVLaS Support for copying the model from a PVC to a workspace wasn't migrated to the e2e pipeline. Is it on purpose? If so, I would propose on removing this part from docs.

pipelines/README.md
```diff
@@ -61,133 +61,159 @@ lightgbm-mushrooms/
...
+  ```bash
+  oc apply -f <downloaddir>/rhoai-edge-build-secret.yml
+  oc secret link pipeline rhoai-edge-build-pull-secret
+  ```
+
+### Setup credentials
+Create a copy of the file(s) below to include the required credentials for accessing any models and image registries required for the Pipeline to run successfully:
+* S3 Storage - [credentials-s3.secret.yaml.template](tekton/aiedge-e2e/templates/credentials-s3.secret.yaml.template) to include your credentials required to access any model stored in S3 then apply it to the server
+* Image Registry - [credentials-image-registry.secret.yaml.template](tekton/aiedge-e2e/templates/credentials-image-registry.secret.yaml.template) to include the username and password with write access to the image repository.
+  This is needed only in case you noted username and password of the robot account. In case you directly downloaded and applied a Kubernetes Secret, this file can be skipped.
```
MarianMacik, line 108: Which is the supported way here? `basic-auth` or `dockerconfigjson` secret? Here I mentioned both but the PR check uses the `basic-auth` method. Probably it doesn't matter as both ways are supported by Tekton.

MarianMacik (commented):
I left a few questions because I wasn't sure @LaVLaS.

Also, in case you would like to have the old pipelines somewhere else (or use `deprecated` instead of  `obsolete`), just tell.

pipelines/README.md
LaVLaS: I think the default Tekton instructions follow the basic-auth route but `dockerconfigjson` is the easier workflow given that quay.io will provide a [kubernetes secret](https://access.redhat.com/documentation/en-us/red_hat_quay/3.3/html/use_red_hat_quay/use-quay-manage-repo#allow-robot-access-user-repo) with the `.dockerconfigjson` file that is supported by OpenShift Pipelines

MarianMacik: OK, I will leave the docs mentioning both and in the PR check we will use the `basic-auth`. If that is OK.

README.md
LaVLaS: Follow-up from an offline conversation: If this is the only reference to `pipelines/tekton/obsolete` in the repo after this change then I think we can safely delete the `pipelines/tekton/obsolete` directory from the repo.

After that this is an LGTM from me

MarianMacik: @LaVLaS I originally thought you wanted to "archive" the old pipelines for some reason, but it's true it was a [subtask](https://issues.redhat.com/browse/RHOAIENG-1881) of the [JIRA](https://issues.redhat.com/browse/RHOAIENG-1578) you closed. Regarding the reference to `pipelines/tekton/obsolete` it's purely on us whether we want to retain this reference in the docs or not. If we want to remove the pipelines completely, I have no issues in doing that, I will just add one more commit on top of this so we have the pipelines archived with the correct documentation in the git history.

LaVLaS: > I will just add one more commit on top of this so we have the pipelines archived with the correct documentation in the git history.

That sounds good. Apologies for the confusion.

LaVLaS:
/lgtm

Looks great! I updated the merge conflict.

PR #199, opened 2024-01-30
RHOAIENG-805: Fix for pipelineRuns failure and check-model-and-contai…

Sara4994:
…nerfile-exists task

## Description
This PR fixes the failure of Pipelines happens when two both(Tensorflow and Bike rentals) run in parallel. [[JIRA]](https://issues.redhat.com/browse/RHOAIENG-805)
Also, it fixes the task 'check-model-and-containerfile-exists' to check the correct modelRelativePath and to fail if incorrect. [[JIRA]](https://issues.redhat.com/browse/RHOAIENG-862)

## How Has This Been Tested?
Mount all the required resources along with both the pipelineRuns.

## To Reproduce the issue:
Paste the commands all in one and then check the PipelineRuns in console.

Alternatively, do

oc create -f tekton/build-container-image-pipeline/aws-env-real.yaml
oc apply -k tekton/build-container-image-pipeline/
oc create -f tekton/build-container-image-pipeline/build-container-image-pipelinerun-tensorflow-housing.yaml
watch the first TaskRun (kserve-download-model) turn green in the console, then paste

oc create -f tekton/build-container-image-pipeline/build-container-image-pipelinerun-bike-rentals.yaml
and observe both the PipelineRuns in the console.

## Merge criteria:


grdryn:
@Sara4994 In the PR description here (and maybe even in the commit message), I'd suggest linking to both jira issues. Additionally, in the "Description" section, you could put information on what the issues were and how this change fixes them. In the "How has this been tested" section, you could put details on how you were reproducing the issues (with the implication that those steps no longer cause the issues).

pipelines/tekton/build-container-image-pipeline/check-model-and-containerfile-exists.yaml
```diff
@@ -20,11 +20,13 @@ spec:
     script: |
       #!/usr/bin/env bash
 
+      set -Eeuo pipefail
+
       # Check model files exist
       if [ -n "$(params.modelRelativePath)" ]; then
-        ls -l model_dir/$(params.modelRelativePath)/$(params.model-name)/
+        ls -l model_dir-$(params.model-name)/$(params.modelRelativePath)$(params.model-name)/
```
grdryn, line 27: Changes will likely be needed in the "copy-model-from-pvc" task for this also, since that assumes `model_dir`. Also then the corresponding code in `pipelines/model-upload-Makefile`.

grdryn (changes requested):
These changes need to be done for the aiedge-e2e pipeline also, since that's the current main one now.

pipelines/model-upload/Makefile
```diff
@@ -20,11 +20,11 @@ create:
 
 # upload model to the pod to the PVC
 	@oc exec local-model-to-pvc-pod -- mkdir workspace/${PVC}/model_dir/ -p
-	@oc cp ${MODEL_PATH} local-model-to-pvc-pod:/workspace/${PVC}/model_dir/${NAME}
+	@oc cp ${MODEL_PATH} local-model-to-pvc-pod:/workspace/${PVC}/model_dir-${NAME}/${NAME}
```
MarianMacik, line 23: I wonder whether this is needed as the fix is needed just for the older pipeline, not aiedge-e2e. We can just write in the docs that the location of the uploaded model is changed if somebody wants to run the old pipeline.

FYI the separate build & test pipelines are being deprecated by #198.

MarianMacik:
> My understanding is that that's just in the example `PipelineRun`, rather than an aspect of the `Pipeline` itself. Users will create their own PipelineRun files (or trigger Pipelines from the UI), where they may specify either `PersistentVolumeClaim` or `VolumeClaimTemplate` to satisfy the workspace requirements for their case.
> 
> An advantage of `VolumeClaimTemplate` is that it avoids any state collisions like this. An advantage of using an existing `PersistentVolumeClaim` is that in a more restricted edge environment, there may only be a set number of pre-provisioned PersistentVolumes, rather than an almost limitless on-demand provisioning that would typically be seen in a cluster hosted in the cloud. So depending on where it's running (among other reasons), a user may choose one or the other.

You are right, I originally thought we would for example support just the `VolumeClaimTemplate` but the example you said makes sense. Then it is definitely worth to split the directories.

Also, is it enough to split just by the model name? Maybe it is just safe to expect that there won't be 2 pipelines running for the same model concurrently?

Sara4994:
> > My understanding is that that's just in the example `PipelineRun`, rather than an aspect of the `Pipeline` itself. Users will create their own PipelineRun files (or trigger Pipelines from the UI), where they may specify either `PersistentVolumeClaim` or `VolumeClaimTemplate` to satisfy the workspace requirements for their case.
> > An advantage of `VolumeClaimTemplate` is that it avoids any state collisions like this. An advantage of using an existing `PersistentVolumeClaim` is that in a more restricted edge environment, there may only be a set number of pre-provisioned PersistentVolumes, rather than an almost limitless on-demand provisioning that would typically be seen in a cluster hosted in the cloud. So depending on where it's running (among other reasons), a user may choose one or the other.
> 
> You are right, I originally thought we would for example support just the `VolumeClaimTemplate` but the example you said makes sense. Then it is definitely worth to split the directories.
> 
> Also, is it enough to split just by the model name? Maybe it is just safe to expect that there won't be 2 pipelines running for the same model concurrently?
> 
> @Sara4994 FYI there are 3 commits now, ideally you would have 2, each covering a separate jira.

@MarianMacik yep the commits are resolved now... thank you :)

pipelines/tekton/aiedge-e2e/tasks/copy-model-from-pvc.yaml
```diff
@@ -11,14 +11,14 @@ spec:
   - name: copy-model-from-pvc
     image: quay.io/opendatahub/kserve-storage-initializer:v0.11
     script: |
-      SOURCE_PATH="$(workspaces.source-workspace.path)/model_dir/$(params.model-name)"
+      SOURCE_PATH="$(workspaces.source-workspace.path)/model_dir-$(params.model-name)/$(params.model-name)"
```
MarianMacik, line 14: I'm sorry for the confusion. Please also make change in the Makefile when uploading the model to PVC, otherwise this will not work.

MarianMacik (commented):
Looks good to me now. I have just one question.

Is it enough to split just by the model name? Maybe it is just safe to expect that there won't be 2 pipelines running for the same model concurrently?

pipelines/tekton/aiedge-e2e/tasks/check-model-and-containerfile-exists.yaml
```diff
@@ -25,17 +25,19 @@ spec:
     script: |
       #!/usr/bin/env bash
 
+      set -Eeuo pipefail
+
       # Check model files exist
       if [ -n "$(params.modelRelativePath)" ]; then
-        ls -l model_dir/$(params.modelRelativePath)/$(params.model-name)/ ;
```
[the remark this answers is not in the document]
MarianMacik: Locally if I do `ls -l /home////dir` and `ls -l /home/dir`, both work and resolve to the same directory. So leaving a slash here should work no matter how many slashes will be added by a user.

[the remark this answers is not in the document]
Sara4994: @grdryn @MarianMacik Yep it works with extra slashes as well, so reverting my change here

grdryn:
> Is it enough to split just by the model name? Maybe it is just safe to expect that there won't be 2 pipelines running for the same model concurrently?

That's a good question, and one that we discussed a little while pairing on this change, and I think it's a good conversation to continue here.

In my opinion, there are a couple of variables to consider:
- How the `buildah-cache` workspace requirement is fulfilled (e.g. PersistentVolumeClaim vs VolumeClaimTemplate)
- The expectation of the user when reusing an existing shared PersistentVolume

As we know, the issue will only occur when the same PV is reused for two PipelineRuns at the same time, and your question highlights that this fix will only solve for the case where two runs are in flight for _different_ model names, and if there are two runs in flight for the same model, then the problem could still occur. So this kind of just reduces the possibility, but it could still occur, right?

Some possible solutions, and their limitations:

1. We could generate some unique part for the directory, rather than using the model name, but I'd expect that that would lead to a different problem, where the volume is filled up over time, since there would be directories for old runs kept indefinitely (at least with VolumeClaimTemplate, old PVCs would get deleted when the corresponding old PipelineRun is deleted).

2. We could document that we only support VolumeClaimTemplate to satisfy the workspace, but that might not be ideal if the build is happening in an edge location that doesn't have a storage provisioner that can create new PVs on demand (maybe there are N PVs pre-provisioned).

3. We could try to implement some sort of concurrency management, where we block a run from continuing if there's already one from the same pipeline in flight (I don't believe Tekton has this ability right now?). I'm not sure what the best approach to that would be, or whether it should be in Tekton itself or outside. Maybe there's some existing pattern we could use for this?

4. We could document the limitation as part of this PR, that only one run for a particular model should be run at a time, if using a shared PV?

WDYT @MarianMacik? Also @LaVLaS you might have better thoughts on this :)

LaVLaS:
There are alot of good conversations about making the model directory unique to support multiple PipelineRuns in the same PVC.  Given that we can't enforce concurrency beyond the k8s scheduler managing multiple Pods requesting access to the same PVC and we can't prevent a user from running parallel pipelines on the same model, we could settle a "globally unique [tekton] identifier" like the PipelineRun uid (`context.pipelineRun.name`) OR PipelineRun name ( `context.pipelineRun.uid`) as the directory name which could be supplied to all tasks as a parameter.  

Would this simplify the initial problem to 
```
mkdir -p model_dir/<model_name>/$(context.pipelineRun.name)
```

grdryn:
We had a group discussion about this today, and our conclusion for now is that we'll continue with the current approach in this pull request (`model-dir_<model name>`) for now (we may do something better in future), but document some things around this:

- We strongly recommend using VolumeClaimTemplate rather than re-using a PersistentVolumeClaim
  - If users have a limited number of PVs, they should set the retention of PipelineRuns low enough that they won't run out
- If you decide to reuse a PersistentVolumeClaim for any of the workspaces, then you shouldn't reuse it for any runs with the same model concurrently, as the runs can interfere with each others' state.

MarianMacik:
@grdryn sounds reasonable. Another option would be to incorporate the uid/name of the pipeline and document that cleaning is up to a user if a reusable PersistentVolumeClaim is used. The idea is that if the management of the PVCs is "manual" by the user, so should be the cleanup, at least for now.

But I am fine with the current approach as well.

grdryn:
@MarianMacik Yes, this is along the lines of what @LaVLaS was originally suggesting. We could do that, but in that case we should provide instructions on cleanup in our docs, _and_ the user has to do something outside the standard flow. IMHO, the easiest thing for us to do right now is to make it clear that our recommendation is VolumeClaimTemplate if you want concurrency, otherwise just make sure you don't run two of the same model concurrently.

grdryn (changes requested):
See comment about documentation changes around limitations that we'd like to have here. Thanks!

Sara4994:
> @MarianMacik Yes, this is along the lines of what @LaVLaS was originally suggesting. We could do that, but in that case we should provide instructions on cleanup in our docs, _and_ the user has to do something outside the standard flow. IMHO, the easiest thing for us to do right now is to make it clear that our recommendation is VolumeClaimTemplate if you want concurrency, otherwise just make sure you don't run two of the same model concurrently.

@grdryn included this instruction as part of readme

pipelines/README.md
```diff
@@ -182,6 +182,7 @@ Be sure to also use the correct config map with the test data.
 # From the root folder where this README is located
 oc create -f tekton/aiedge-e2e/aiedge-e2e.pipelinerun.yaml
 ```
+Note: It is recommended to use VolumeClaimTemplate if the usecase is to run multiple pipelineruns concurrently, otherwise just make sure you don't run two of the same model concurrently.
```
grdryn, line 185:
```suggestion
> [!IMPORTANT]
> Since the `buildah-cache` workspace is used to share data between TaskRuns in a PipelineRun, a PersistentVolumeClaim type VolumeSource is required to fulfill it properly.
> We strongly recommend that this is fulfilled using the `volumeClaimTemplate` approach, rather than the `persistentVolumeClaim` approach.
> If you must use the `persistentVolumeClaim` approach to re-use an existing PersistentVolumeClaim, then you will likely hit issues if two PipelineRuns for the same model name are executed concurrently (and possibly other corner cases).
> See the Tekton documentation around [Using PersistentVolumeClaims as VolumeSource](https://tekton.dev/docs/pipelines/workspaces/#using-persistentvolumeclaims-as-volumesource).
```

See https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts

grdryn (approved):
/lgtm

Thanks @Sara4994!

PR #202, opened 2024-01-31
RHOAIENG-183: Use ODH OpenVino image, and other fixes

grdryn:
JIRA: https://issues.redhat.com/browse/RHOAIENG-183

## Description

- Use ODH OpenVino image in base Containerfile
The source for that image is available at https://github.com/opendatahub-io/openvino_model_server
I also squashed a coupld of RUN commands into one in that Containerfile.
- Use a consistent default for MODEL_DIR in Containerfiles
I don't know whether it's the best value, but the previous on in the OpenVino Containerfile seemed too specific to the tensorflow example model.
- This makes it easier to use the same example curl command to test that the image built correctly, whether from our Pipeline or using the example `podman run` in this README.
  - Use ODH OpenVino base image
  - Use same REST port as pipeline Containerfiles use (8080)
  - Use onnx-mnist as the name rather than onnx_mnist
  - Use Podman in examples 🦭
- Add OpenVino PipelineRun example for aiedge-e2e
This also renames the pre-existing one to make it clear that it's for the bike-rentals example, and updates the README to mention both.

## How Has This Been Tested?

- The included PipelineRun file for the tensorflow-housing example was created to test this change. Additionally, I ran the secondary example models locally with Podman.

## Merge criteria:


pipelines/models/tensorflow-facedetection/README.md
```diff
@@ -15,7 +15,7 @@ docker run -d -u $(id -u):$(id -g) --rm -v ${PWD}/models:/model:Z -p 9000:9000 o
 git clone https://github.com/openvinotoolkit/model_server.git
 cd model_server/demos/face_detection/python
 # Patch the requirements file, because tensorflow 2.11.0 is not working
-sed -i 's/2.11.0/2.13.0/' client_requirements.txt
+sed -i 's/tensorflow-serving-api==2.11.0/tensorflow-serving-api==2.13.0/' ../../common/python/requirements.txt
 python -m venv .venv
 source .venv/bin/activate
 pip install -r ../../common/python/requirements.txt
```
grdryn, line 21: I actually couldn't get this particular test request to work...I had issues with Python dependencies.

pipelines/containerfiles/Containerfile.openvino.mlserver.mlflow
```diff
@@ -17,12 +17,16 @@ RUN mkdir /models && chown ovms:ovms /models
 
 # CHANGE THIS LINE TO MATCH YOUR MODEL
 COPY --chown=ovms:ovms $MODEL_DIR /models/1
-# Not deleting this file causes a bug
-RUN if echo "${MODEL_DIR}" | grep -q "tf2model"; then rm -f /models/1/fingerprint.pb; fi
 
-RUN chmod o+rwX /models/1
-# https://docs.openshift.com/container-platform/4.13/openshift_images/create-images.html#use-uid_create-images
-RUN chgrp -R 0 /models/1 && chmod -R g=u /models/1
+RUN <<EOF
```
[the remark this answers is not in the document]
grdryn: @davidesalerno in my understanding, every RUN instruction in a Dockerfile results in an additional layer in the resulting image. Sometimes there are good reasons to split RUN instructions into separate layers, but if there isn't, I think it's a better practice to keep them in a single layer. This can avoid size bloat in cases where, for example, two commands get run and the second removes some temporary files that the first one creates. While that example doesn't apply to this particular case right now, I think it's a reasonable idea to group these commands into a single RUN, so that it might be more obvious to people making changes in the future, that they can just put more commands into it too, if it makes sense.

Counter question, to make sure I'm not missing anything: are there reasons that you can think of that we _shouldn't_ group these commands into a single RUN like this?

Also, with reference to the "rearrange" part of your question: they're not rearranged into a different order, they will all be executed in the same order that they were in previously.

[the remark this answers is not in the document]
MarianMacik: Nice, didn't even know this is possible.

pipelines/models/onnx-mnist/README.md
```diff
@@ -6,17 +6,18 @@ This model predicts handwritten digits using a convolutional neural network (CNN
...
 ```
 ## Credits
-- https://github.com/onnx/models/tree/main/archive/vision/classification/mnist
\ No newline at end of file
+- https://github.com/onnx/models/tree/main/archive/vision/classification/mnist
```
MarianMacik, line 23: This has been moved.
```suggestion
- https://github.com/onnx/models/tree/main/validated/vision/classification/mnist
```

pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
```diff
@@ -30,6 +30,10 @@ spec:
   - name: modelRelativePath
     type: string
     description: "Location of the model within the context of the source location.  Leave blank if the model files are at the root of the source location."
+  - name: model-dir
+    type: string
+    description: Directory below <modelRelativePath>/<model-name>/ to be used as MODEL_DIR in image build.
```
MarianMacik, line 35: We need to revisit this in a separate PR/Jira as effectively we now have 2 relative paths :) But I think an issue for that already exists, I will check.

pipelines/tekton/aiedge-e2e/aiedge-e2e.tensorflow-housing.pipelinerun.yaml
```diff
@@ -0,0 +1,59 @@
...
+  - name: gitServer
+    value: https://github.com
+  - name: gitOrgName
+    value: opendatahub-io
+  - name: gitRepoName
+    value: ai-edge
+  - name: containerfileRelativePath
+    value: pipelines/containerfiles/Containerfile.openvino.mlserver.mlflow
+  - name: fetch-model
+    value: "git"
```
MarianMacik, line 24:
```suggestion
    value: "s3"
```
Or overwrite this in the pipeline tests.
Currently we test S3 using both examples as this is more important than Git use case AFAIK. There is a plan to include also Git testing, of course, but just for now I would stick with the git.

test/shell-pipeline-tests/openvino-tensorflow-housing/pipelines-test-openvino-tensorflow-housing.sh
```diff
@@ -35,13 +35,7 @@ oc apply -k "$AIEDGE_E2E_PIPELINE_DIR_PATH"/
...
 ## prepare parameters
 AIEDGE_E2E_PIPELINE_OVERRIDDEN_PATH="$AIEDGE_E2E_PIPELINE_DIR_PATH"/aiedge-e2e.pipelinerun-overridden.yaml
-cp "$AIEDGE_E2E_PIPELINE_DIR_PATH"/aiedge-e2e.pipelinerun.yaml "$AIEDGE_E2E_PIPELINE_OVERRIDDEN_PATH"
-sed -i "s|value: rhoai-edge-models|value: rhoai-edge-models-ci|" "$AIEDGE_E2E_PIPELINE_OVERRIDDEN_PATH"
-sed -i "s|value: bike-rentals-auto-ml|value: tensorflow-housing|" "$AIEDGE_E2E_PIPELINE_OVERRIDDEN_PATH"
-sed -i "s|Containerfile.seldonio.mlserver.mlflow|Containerfile.openvino.mlserver.mlflow|" "$AIEDGE_E2E_PIPELINE_OVERRIDDEN_PATH"
-sed -i "s|value: pipelines/models/|value: \"\"|" "$AIEDGE_E2E_PIPELINE_OVERRIDDEN_PATH"
-sed -i "s|\"invocations\"|\"v1/models/tensorflow-housing/versions/1:predict\"|" "$AIEDGE_E2E_PIPELINE_OVERRIDDEN_PATH"
-sed -i "s|name: bike-rentals-test-data|name: tensorflow-housing-test-data|" "$AIEDGE_E2E_PIPELINE_OVERRIDDEN_PATH"
+cp "$AIEDGE_E2E_PIPELINE_DIR_PATH"/aiedge-e2e.tensorflow-housing.pipelinerun.yaml "$AIEDGE_E2E_PIPELINE_OVERRIDDEN_PATH"
```
MarianMacik, line 38:
```suggestion
cp "$AIEDGE_E2E_PIPELINE_DIR_PATH"/aiedge-e2e.tensorflow-housing.pipelinerun.yaml "$AIEDGE_E2E_PIPELINE_OVERRIDDEN_PATH"
sed -i "s|value: rhoai-edge-models|value: rhoai-edge-models-ci|" "$AIEDGE_E2E_PIPELINE_OVERRIDDEN_PATH"
sed -i "s|value: pipelines/models/|value: \"\"|" "$AIEDGE_E2E_PIPELINE_OVERRIDDEN_PATH"
```
+ possibly git to s3 or you can do it in the pipelinerun directly.

pipelines/tekton/aiedge-e2e/aiedge-e2e.tensorflow-housing.pipelinerun.yaml
grdryn: @MarianMacik what's the value of testing S3 in both examples? Wouldn't the fact that it's used in one cover that pipeline task, and using git in the other one be additional coverage for the git clone task? My assumption was that they were of equal/similar importance.

pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
grdryn: Yes, as mentioned in the commit message for this one, it's kind of a hack because of something that was hardcoded in the Containerfile. IMHO it's a symptom of having one uber-generic pipeline to try to cover all types of models and model servers.

test/shell-pipeline-tests/openvino-tensorflow-housing/pipelines-test-openvino-tensorflow-housing.sh
grdryn: As mentioned in the other comment (which I think is related to this, right?), I think having one test use a model from S3 and the other one use an example from Git would be better than having both from S3.

MarianMacik: @grdryn Yes, possibly. When I was implementing these tests I didn't see into it that much. The question is whether we need to test both example models on S3 and Git or not. Can it happen that one example model will run fine when fetched from S3 but not from Git or vice versa?

One thing which may theoretically occur is that we somehow move model/rename it in this Git repo and we forget to copy the changes to the S3 or vice versa as S3 contains a copy of these models. However, this wouldn't be an issue of our pipeline, of course. I am just thinking out loud whether we want to have the full coverage of all combinations or not.

If we are fine with having one for S3 and the other for Git then you can set relative path to `""` for the s3 seldonio-bike-rentals and remove that sed line from the test.

grdryn: > The question is whether we need to test both example models on S3 and Git or not. Can it happen that one example model will run fine when fetched from S3 but not from Git or vice versa?

In my opinion, the example models aren't the crucial thing that we want to verify the correctness of, it's the Pipeline and its Tasks. I don't think we would get any additional value from testing each model with each download method. By running one instance of the pipeline (for any model) with the S3 method, and another instance of the pipeline (for any model) with the Git method, then we're getting good coverage of the different Pipeline variations.
It's also good that the two examples we're using in the tests use different model servers, which tests the different Containerfiles that we have for those.

> One thing which may theoretically occur is that we somehow move model/rename it in this Git repo and we forget to copy the changes to the S3 or vice versa as S3 contains a copy of these models. However, this wouldn't be an issue of our pipeline, of course. I am just thinking out loud whether we want to have the full coverage of all combinations or not.

At that point, I think it's almost like what we'd be testing is the structure of the `rhoai-edge-models-ci` S3 bucket (which isn't an output of this project), rather than the pipeline or the example, right? We don't want to "overfit the data" (to borrow a term from machine learning) – as long as users make a PipelineRun that is correct for their model and where it's stored, our Pipeline should succeed. It shouldn't need to be completely like the structure or naming of our examples, as long as they can fit the Pipeline to it by configuring the parameters that we've made available.

> If we are fine with having one for S3 and the other for Git then you can set relative path to "" for the s3 seldonio-bike-rentals and remove that sed line from the test.

Done now, thanks! :+1:

MarianMacik: Yes, you are right, I think this way it's both sufficient and more efficient. If it was a very short test it would be doable but it doesn't make sense to test every possible combination with such long-running tests.

Makefile
```diff
@@ -35,6 +35,6 @@ GO=go
 GOFLAGS=""
 
 test:
-	@(./test/shell-pipeline-tests/openvino-bike-rentals/pipelines-test-openvino-bike-rentals.sh)
-	@(./test/shell-pipeline-tests/tensorflow-housing/pipelines-test-tensorflow-housing.sh)
+	@(./test/shell-pipeline-tests/seldon-bike-rentals/pipelines-test-seldon-bike-rentals.sh)
+	@(./test/shell-pipeline-tests/openvino-tensorflow-housing/pipelines-test-openvino-tensorflow-housing.sh)
 	@(cd test/e2e-tests/tests && ${GO} test)
```
MarianMacik, line 40: There is a conflict after the previous PR was merged.

MarianMacik (commented):
Looks good to me. I think after the conflict is solved, I can give an approval. Unless we want to wait for others, of course.

Makefile
grdryn: Fixed, thanks :+1:

MarianMacik:
@davidesalerno @grdryn are we fine with approving this? If so, I can go ahead.

MarianMacik (approved):
Thanks!

PR #204, opened 2024-02-07
Fix the ApplicationSet clusterDefinition generator

devguyio:
## Description

In the `ApplicationSet`, the name of the `ConfigMap` referenced in the cluster decision resource generator was incorrect so syncing of the applications was broken. 

## How Has This Been Tested?

This was tested by following the README on a rosa hub cluster and an OpenShift local edge cluster.

## Merge criteria:

### DIFF
--- a/acm/registration/near-edge/base/near-edge.yaml
+++ b/acm/registration/near-edge/base/near-edge.yaml
@@ -65,7 +65,7 @@ metadata:
 spec:
   generators:
     - clusterDecisionResource:
-        configMapRef: ocm-placement-generator
+        configMapRef: acm-placement
         labelSelector:
           matchLabels:
             cluster.open-cluster-management.io/placement: placement


acm/registration/near-edge/base/near-edge.yaml
```diff
@@ -65,7 +65,7 @@ metadata:
 spec:
   generators:
     - clusterDecisionResource:
-        configMapRef: ocm-placement-generator
+        configMapRef: acm-placement
```
LaVLaS, line 68: Is this a ConfigMap that is provide with ACM `multiclusterhub` install

devguyio: I believe so yes, I installed ACM from the UI but this ConfigMap is provided by ACM definitely and I assume it gets created when the `MultiClusterHub` object is created.

grdryn: I can see that it's immediately added to the `openshift-gitops` namespace on a cluster that has the ACM hub installed, when that namespace is created & an ArgoCD CR installed into it. The `ocm-placement-generator` one is not.

Trying to piece together why this worked for me with `ocm-placement-generator`, I can only assume that I was basing it on the example in the open-cluster-management-io repo, which includes [this file](https://github.com/open-cluster-management-io/argocd-pull-integration/blob/main/example/hub/ocm-placement-generator-cm.yaml) and I either had that in the namespace or I originally had included it as part of the manifests in this repo and then removed it before the final version that got merged

I can't explain how it would have worked for whoever reviewed/tested my PR :upside_down_face:

My question now is whether the change in this PR is the fix we want, or whether we should create a ConfigMap called `ocm-placement-generator` (or something) as part of our `oc apply -k`, so that it should work for both ACM and the upstream OCM? Since the values in it will be the same either way, I don't see a major problem with us holding a version of it in our configs. What do y'all think?

MarianMacik: @grdryn so did you try it using the community version (OCM) and that is why it wasn't created when @devguyio tried it with ACM? Do I understand it?

grdryn: @MarianMacik I'm not sure if I ever tried it with OCM rather than ACM, but I would have used the example that I linked from the OCM argocd-pull-integration repo that I linked above. This would have created that ConfigMap and then it likely wasn't removed & I kept relying on it because it kept working on that cluster.

What I'm saying in my last paragraph is that _maybe_ we should create our own instance of that ConfigMap so that we know the name will always be correct. We're asking for input/feedback from OCM folks on that idea now.

grdryn (approved):
/lgtm

Since this fixes a broken thing, it's worth getting merged ASAP. We can still change it to the alternative approach being discussed at any time if we decide to.

I've verified in my environment that this is broken without this change, and that this change fixes the issue. Thanks @devguyio!

acm/registration/near-edge/base/near-edge.yaml
grdryn: > So if we create that config map ourselves, then the placementdecision configured in it would still need to be created by ourselves? Or is that one prepopulated automatically by clusters?

That ConfigMap doesn't actually contain a PlacementDecision or even a reference to a specific one. My understanding is that it's just boilerplate needed to get the ArgoCD ApplicationSet generator stuff (which is very generic) to be able to understand how to make Applications based on OCM/ACM PlacementDecisions.

We already do create the Placements required (we have to decide for each application (e.g. the bike-rentals one) which clusters it should be deployed to), which will cause OCM/ACM to create the corresponding PlacementDecisions, which then get used by ArgoCD through the ApplicationSet spec.generators section.

devguyio: I think there's no need to create an extra ConfigMap. The intention of the ConfigMap is to specify the GVK of the placement duck type that will be used by the ArgoCD instance. So we should juts have one ConfigMap per GVK. Since that's the correct GVK, then we're using it correctly as intended.

grdryn: @devguyio the case where I'm suggesting that we might want to create the ConfigMap ourselves is if we happen to want to also allow this to work with OCM, and if it happens to uses a different name for the ConfigMap or doesn't include an equivalent ConfigMap at all. If we created it ourselves, we wouldn't have to care whether it's there for either OCM or ACM, or if they used the same or different names.

devguyio: > @devguyio the case where I'm suggesting that we might want to create the ConfigMap ourselves is if we happen to want to also allow this to work with OCM, and if it happens to uses a different name for the ConfigMap or doesn't include an equivalent ConfigMap at all. If we created it ourselves, we wouldn't have to care whether it's there for either OCM or ACM, or if they used the same or different names.

I see your point, I'm just saying since we're not there yet then IMHO we shouldn't duplicate the config map as we'd be doing that for a requirement that we currently don't have. When we have that requirement, a decent effort will be done to see how to make it work with both and what other changes are gonna be needed, which might be just this one.

PR #205, opened 2024-02-14
RHOAIENG-3233 - Include events and deployment logs in the shell-pipel…

MarianMacik:
…ine-tests
**JIRA:** [RHOAIENG-3233](https://issues.redhat.com/browse/RHOAIENG-3233)
## Description
This adds:
* Archiving events of the namespace at the end of the scenario (events.txt)
* Archiving deployment and its pod's log (deployments.txt and deploymentLogs.txt)
  * This required including the `upon-end` parameter in the pipeline run even though it has a default value so it can be easily overwritten using sed to `keep` the deployment running, otherwise logs would be lost or obtained just in case the pod termination would be long enough and end only after the pipeline finishes. Without this change it would need more complicated manipulation of the yaml file using tools such as `yq` etc.

## How Has This Been Tested?
Locally + using a PR check.

## Merge criteria:

### DIFF
--- a/pipelines/tekton/aiedge-e2e/aiedge-e2e.bike-rentals.pipelinerun.yaml
+++ b/pipelines/tekton/aiedge-e2e/aiedge-e2e.bike-rentals.pipelinerun.yaml
@@ -32,6 +32,8 @@ spec:
     value: "invocations"
   - name: target-imagerepo
     value: rhoai-edge
+  - name: upon-end
+    value: "delete"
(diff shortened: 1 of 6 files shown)


grdryn (approved):
I've tested one of the pipelineruns manually and it succeeded as expected. I've also taken a look at the new files created by the shell tests in the artifacts from the successful run from this PR, and they look like a very useful addition.

Thanks @MarianMacik!

PR #206, opened 2024-02-15
Makefile: Add support to include vars from local env var file

LaVLaS:
## Description
Adds support for passing Make variables via a local env var file instead of on the command line

## How Has This Been Tested?
Successfully ran `make` against `go-test` and `go-test-setup` without any variables on the CLI

## Merge criteria:
CI runs and local `make` execution works without any issues

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -84,3 +84,33 @@ make -s -e GIT_REPO_URL="https\://github.com/opendatahub-io/ai-edge" \
      CUSTOM_APP_NAMESPACE=my-test-namespace \
      test-acm-bike-rental-app-generate  | oc apply -f -
 ```
+
+## Using a local.vars.mk file to override Makefile variables for your development environment
+To support the ability for a developer to customize the Makefile execution for their development environment, you can create a `local.vars.mk` file in the root of this repo to specify custom values matching your environment.  
+
+```
+$ cat local.vars.mk
+AWS_SECRET_ACCESS_KEY=MYSECRETACCESSKEYAWS
+AWS_ACCESS_KEY_ID=a1b2c3d4e5f6g7h8i9j0abcdefghijklmnopqrstuv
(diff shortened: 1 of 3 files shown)


DEVELOPMENT.md
```diff
@@ -84,3 +84,33 @@ make -s -e GIT_REPO_URL="https\://github.com/opendatahub-io/ai-edge" \
...
+```
+$ cat foo-storage.local.vars.mk
+AWS_SECRET_ACCESS_KEY=MYSECRETACCESSKEYFOO
+AWS_ACCESS_KEY_ID=a1b2c3d4e5f6g7h8i9j0abcdefghijklmnopqrstuv
+S3_REGION=us-west-4
+S3_ENDPOINT=https://s3.foo-object-storage.com
+IMAGE_REGISTRY_USERNAME=my+robot+username+foo
+IMAGE_REGISTRY_PASSWORD=<IMAGE-REGISTRY-PASSWORD>
+
+$ make MAKE_ENV_FILE=foo-storage.local.vars.mk
```
grdryn, line 114:
```suggestion
$ make -e MAKE_ENV_FILE=foo-storage.local.vars.mk
```

I think the `-e` would be needed here, to force overriding the value specified in the Makefile, right?

grdryn:
An alternative to consider might be [direnv](https://direnv.net/).

MarianMacik (commented):
Looks good to me.

DEVELOPMENT.md
MarianMacik: AFAIK command line variables override variables defined in the Makefile. See https://www.gnu.org/software/make/manual/html_node/Overriding.html

grdryn: Ah yes, my mistake: the `-e` would only be needed if `MAKE_ENV_FILE` was an environment variable, but it's not.

grdryn (approved):
/lgtm

PR #207, opened 2024-02-15
Heredoc in containerfile is supported only since 1.33.0

MarianMacik:
## Description
* Switches heredoc to a one-liner so we have just one layer. Heredoc is only compatible in [Buildah 1.33.0](https://buildah.io/releases/#buildah-version-1330-release-announcement) but the latest rhel8 image has just 1.31.
* Updates the buildah image to the [latest](https://catalog.redhat.com/software/containers/rhel8/buildah/5dca3d76dd19c71643b226d5?image=659525ba18d1f0c9c2d52ba6&architecture=amd64) one.

## How Has This Been Tested?
Locally, PR check currently doesn't run the containerfiles from the same branch. I will investigate whether it is possible to obtain the PR's author and fork.

## Merge criteria:

### DIFF
--- a/pipelines/containerfiles/Containerfile.openvino.mlserver.mlflow
+++ b/pipelines/containerfiles/Containerfile.openvino.mlserver.mlflow
@@ -18,15 +18,8 @@ RUN mkdir /models && chown ovms:ovms /models
 # CHANGE THIS LINE TO MATCH YOUR MODEL
 COPY --chown=ovms:ovms $MODEL_DIR /models/1
 
-RUN <<EOF
-    # Not deleting this file causes a bug
-    if echo "${MODEL_DIR}" | grep -q "tf2model"; then rm -f /models/1/fingerprint.pb; fi
-
-    chmod o+rwX /models/1
-
-    # https://docs.openshift.com/container-platform/4.13/openshift_images/create-images.html#use-uid_create-images
-    chgrp -R 0 /models/1 && chmod -R g=u /models/1
-EOF
+                                # Not deleting this file causes a bug                                                # https://docs.openshift.com/container-platform/4.13/openshift_images/create-images.html#use-uid_create-images
+RUN if echo "${MODEL_DIR}" | grep -q "tf2model"; then rm -f /models/1/fingerprint.pb; fi && chmod o+rwX /models/1 && chgrp -R 0 /models/1 && chmod -R g=u /models/1
 
 EXPOSE $REST_PORT $GRPC_PORT
 
--- a/pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
+++ b/pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
@@ -208,7 +208,7 @@ spec:
     - name: IMAGE
       value: image-registry.openshift-image-registry.svc:5000/$(context.pipelineRun.namespace)/$(params.model-name):$(params.model-version)
     - name: BUILDER_IMAGE
-      value: registry.redhat.io/rhel8/buildah@sha256:0a86ecbdfbe86e9d225b7fe4b090a0dd6d323f8afdfdf2bd933ff223ddb53320
+      value: registry.redhat.io/rhel8/buildah@sha256:f5521de9445a04054c386707b774d7d534556c03b29ba4c91cfddfa979cf761c
     - name: STORAGE_DRIVER
       value: vfs
     - name: DOCKERFILE


pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
```diff
@@ -208,7 +208,7 @@ spec:
     - name: IMAGE
       value: image-registry.openshift-image-registry.svc:5000/$(context.pipelineRun.namespace)/$(params.model-name):$(params.model-version)
     - name: BUILDER_IMAGE
-      value: registry.redhat.io/rhel8/buildah@sha256:0a86ecbdfbe86e9d225b7fe4b090a0dd6d323f8afdfdf2bd933ff223ddb53320
+      value: registry.redhat.io/rhel8/buildah@sha256:f5521de9445a04054c386707b774d7d534556c03b29ba4c91cfddfa979cf761c
```
grdryn, line 211: /me wonders if we should be using UBI rather than RHEL, and 9 instead of 8 (not that it would fix this actual issue)

pipelines/containerfiles/Containerfile.openvino.mlserver.mlflow
```diff
@@ -18,15 +18,8 @@ RUN mkdir /models && chown ovms:ovms /models
...
-RUN <<EOF
-    # Not deleting this file causes a bug
-    if echo "${MODEL_DIR}" | grep -q "tf2model"; then rm -f /models/1/fingerprint.pb; fi
-
-    chmod o+rwX /models/1
-
-    # https://docs.openshift.com/container-platform/4.13/openshift_images/create-images.html#use-uid_create-images
-    chgrp -R 0 /models/1 && chmod -R g=u /models/1
-EOF
+                                # Not deleting this file causes a bug                                                # https://docs.openshift.com/container-platform/4.13/openshift_images/create-images.html#use-uid_create-images
```
grdryn, line 21:
```suggestion
# Not deleting this file causes a bug
# https://docs.openshift.com/container-platform/4.13/openshift_images/create-images.html#use-uid_create-images
```

MarianMacik: @grdryn Here I opted for matching the comments to the actual part of that line, so it's easier to understand what belongs to which part. Or do you think it would be understandable having it at the start of the line?

pipelines/containerfiles/Containerfile.openvino.mlserver.mlflow
```diff
@@ -18,15 +18,8 @@ RUN mkdir /models && chown ovms:ovms /models
...
+RUN if echo "${MODEL_DIR}" | grep -q "tf2model"; then rm -f /models/1/fingerprint.pb; fi && chmod o+rwX /models/1 && chgrp -R 0 /models/1 && chmod -R g=u /models/1
```
LaVLaS, line 22: Does anyone remember why we have to remove `fingerprint.pb` from the file?

Is setting `chmod o+rwX` required?  This would open permissions to anyone user with access?

Not a reason to block this PR as everything can be addressed in a follow-up PR

MarianMacik: @LaVLaS I don't know.

pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
LaVLaS: I agree on the switch to ubi9 for parity with upstream implementation

Without regular automation to update the digest hash, we should switch to the tag.

We can follow up in other PRs to address the changes

MarianMacik: OK, I will create a jira for that.

pipelines/containerfiles/Containerfile.openvino.mlserver.mlflow
grdryn: It's fine :+1:

pipelines/containerfiles/Containerfile.openvino.mlserver.mlflow
grdryn: Good questions @LaVLaS...we should see if we can simplify things here. I'm not sure what the answers currently are.

pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
MarianMacik: Created [RHOAIENG-3359](https://issues.redhat.com/browse/RHOAIENG-3359)

PR #211, opened 2024-02-20
RHOAIENG-2758: Accept one or more target image tags as a param

grdryn:
JIRA: https://issues.redhat.com/browse/RHOAIENG-2758

## Description
Before this change, the image registry, image name, and image tag, were all hardcoded or generated by the pipeline, rather than customizable through parameters.

This change makes it possible to specify one or more targets as full image tag references.

Accepting more than just a single image tag target allows for both uniqueness and meaning. The simplest example scenario is where there's a primary tag that's unique, which has the model version and pipelinerun id; then a secondary tag with just the model version, which tracks the current blessed image for that model version. Quay (and possibly other registries) even has a UI feature for visualising that different tags actually point to the same manifest.

It's still possible to use other parameters to generate a name just like before, as demonstrated in the example PipelineRun files.

## How has this been tested
- Pipelines run manually, using both example PipelineRuns
- `make go-test` passes
- Verified that this works with the following image registries:
  - [x] quay.io
  - [x] docker.io
  - [x] internal registry in OpenShift cluster

## Merge criteria:

### DIFF
--- a/pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
+++ b/pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
@@ -127,7 +127,7 @@ spec:
       value: /model_dir-$(params.model-name)/
     workspaces:
     - name: output
-      workspace: buildah-cache
+      workspace: build-workspace-pv
     when:
     - input: "$(params.fetch-model)"
       operator: in
@@ -146,7 +146,7 @@ spec:
       name: kserve-download-model
     workspaces:
     - name: workspace
-      workspace: buildah-cache
+      workspace: build-workspace-pv
     - name: s3-secret
       workspace: s3-secret
     when:
@@ -167,7 +167,7 @@ spec:
       name: git-clone
     workspaces:
     - name: output
-      workspace: buildah-cache
+      workspace: build-workspace-pv
     - name: basic-auth
       workspace: git-basic-auth
   - name: check-model-and-containerfile-exists
@@ -187,7 +187,7 @@ spec:
       name: check-model-and-containerfile-exists
     workspaces:
     - name: workspace
-      workspace: buildah-cache
+      workspace: build-workspace-pv
   - name: create-imagestream
     params:
     - name: SCRIPT
@@ -227,7 +227,7 @@ spec:
       name: buildah
     workspaces:
     - name: source
-      workspace: buildah-cache
+      workspace: build-workspace-pv
   - name: test-container-deploy
     params:
     - name: SCRIPT
@@ -356,7 +356,7 @@ spec:
     - stop-deployment
     workspaces:
     - name: images-url
-      workspace: buildah-cache
+      workspace: build-workspace-pv
   - name: skopeo-copy
     params:
     - name: srcTLSverify
@@ -370,9 +370,9 @@ spec:
       name: skopeo-copy
     workspaces:
     - name: images-url
-      workspace: buildah-cache
+      workspace: build-workspace-pv
   workspaces:
-  - name: buildah-cache
+  - name: build-workspace-pv
   - name: s3-secret
   - name: git-basic-auth
     optional: true
--- a/pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
+++ b/pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
@@ -51,9 +51,9 @@ spec:
   - default: $(context.pipelineRun.namespace)
     name: target-namespace
     type: string
-  - name: target-image-tag-reference
-    type: string
-    description: "The fully qualified image tag reference to be used for the final image push. E.g. registry.example.com/my-org/ai-model:1.0-1"
+  - name: target-image-tag-references
+    type: array
+    description: "An array of fully qualified image tag references to be used for the final image push. E.g. registry.example.com/my-org/ai-model:1.0-1"
   - name: upon-end
     type: string
     description: "Action to perform on the k8s deployment created to test the model container image. Valid values in [delete, keep, stop]"
@@ -89,9 +89,9 @@ spec:
   - name: internal-registry-url
     description: The tag where the model container image was pushed to in the internal registry (e.g. image-reg...svc:5000/rhoai-models/ai-model:1-232)
     value: $(tasks.build-container.results.IMAGE_URL)
-  - name: target-image-tag-reference
+  - name: target-image-tag-references
     description: The fully qualified image reference that the image was pushed to (e.g. registry.example.com/my-org/ai-model:1.0-1)
-    value: $(tasks.retrieve-build-image-info.results.target-image-tag-reference)
+    value: $(tasks.retrieve-build-image-info.results.target-image-tag-references)
   - name: internal-image-url
     description: The url of the image in the internal registry
     value: $(tasks.retrieve-build-image-info.results.internal-image-url)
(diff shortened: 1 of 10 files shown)


pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
```diff
@@ -352,17 +349,16 @@ spec:
...
-    - name: target-registry-url
-      value: quay.io/$(params.target-imagerepo)/$(params.model-name)
+    - name: target-image-tag-references
+      value: ["$(params.target-image-tag-references[*])"]
     runAfter:
     - test-model-rest-svc
     - stop-deployment
+    workspaces:
+    - name: images-url
+      workspace: buildah-cache
```
grdryn, line 359: Maybe this should be a different workspace?

pipelines/tekton/aiedge-e2e/tasks/retrieve-build-image-info.task.yaml
```diff
@@ -15,18 +15,24 @@ spec:
...
-    type: string
+  - name: target-image-tag-references
+    type: array
+  workspaces:
+  - name: images-url
   steps:
   - name: get-image-sha
     image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
+    args:
+      - "$(params.target-image-tag-references[*])"
```
grdryn, line 26: Can't reference array param directly in script, so need to convert to `args`, then use `$@` in script.

pipelines/tekton/aiedge-e2e/tasks/retrieve-build-image-info.task.yaml
```diff
@@ -40,7 +46,25 @@ spec:
...
+  - name: build-urls-txt
+    image: registry.access.redhat.com/ubi9/ubi-micro
+    args:
+      - "$(params.target-image-tag-references[*])"
+    script: |
+      #!/usr/bin/env bash
+
+      set -Eeuo pipefail
+
+      export URLTXT=$(workspaces.images-url.path)/url.txt
```
grdryn, line 59: skopeo-copy task looks for this file in its workspace if the source and destination parameters are empty. This is what allows pushing to more than one tag from the single taskrun.

pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
```diff
@@ -51,9 +51,9 @@ spec:
   - default: $(context.pipelineRun.namespace)
     name: target-namespace
     type: string
-  - name: target-imagerepo
-    type: string
-    description: "The image registry name where the model container for the final image push. Currently supports using quay.io/<target-imagerepo>"
+  - name: target-image-tag-references
+    type: array
+    description: "An array of fully qualified image tag references to be used for the final image push. E.g. registry.example.com/my-org/ai-model:1.0-1"
```
grdryn, line 56: This could maybe have a default value of something like:
```
- image-registry.openshift-image-registry.svc:5000/$(context.pipelineRun.namespace)/$(params.model-name):$(params.model-version)-$(context.pipelineRun.uid)
- image-registry.openshift-image-registry.svc:5000/$(context.pipelineRun.namespace)/$(params.model-name):$(params.model-version)-production
```

This could be useful for automated testing, as there wouldn't need to be any registry/credentials specified. Thoughts?

pipelines/tekton/aiedge-e2e/tasks/retrieve-build-image-info.task.yaml
devguyio: Maybe add that as a comment above the export line?

devguyio (commented):
Nice work! Left a few comments but it'd be great if you can also update the pipelines/README with the new parameters and how to use it.

pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
grdryn: Done (well, I've added one of them as a default)

test/e2e-tests/support/options.go
```diff
@@ -55,3 +57,14 @@ func setOptions() (*Options, error) {
 
 	return options, nil
 }
+
+func parseImageTagsJSON(imageTagsJSON string) ([]string, error) {
+	var imageTags []string
+
+	err := json.Unmarshal([]byte(imageTagsJSON), &imageTags)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to unmarshall image tags JSON string (%v) into []string: %w", imageTagsJSON, err)
```
jackdelahunt, line 66: I think error messages are generally lower-case

pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
grdryn: ~@devguyio can you explain what you mean a little more here? What I'm thinking is that it's already being used as more than a buildah-cache, and especially this PR is adding something that's not related to the image build.~ The `skopeo-copy` explicitly needs a file called `url.txt` at the root of the workspace, and I guess at this stage in the pipeline, we shouldn't need to care if something had already put a file with the same name there, so I'll just put a `rm -f url.txt` in the script before appending to it (ugh, actually I'll need to do that anyway, because if the workspace is a reused PV across runs, then we'd be appending to it continuously, and pushing over existing tags...). 

Actually, I think we _at least_ need to rename this workspace from `buildah-cache` to something else, since _buildah isn't caching in this pipeline!_ (in the version (0.6) of the Task that we get from OpenShift Pipelines currently, there's a hardcoded `--no-cache` (the `--no-cache is removed in version 0.7, but that's to allow remote/distributed caching with `--cache-from`/`--cache-to` which that doesn't happen by default, and we don't do it, and it wouldn't be using this workspace to do the caching anyway even if we did).

I think maybe I'll just rename it to "build-workspace", LMK if you think it should be something different.

jackdelahunt:
Tested with `make go-test`, both pipelines passed :+1:

pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
```diff
@@ -51,9 +51,12 @@ spec:
...
+    default:
+    # This default is just an example to show how it can be used to push to the default internal registry in OpenShift
+    - image-registry.openshift-image-registry.svc:5000/$(context.pipelineRun.namespace)/$(params.model-name):$(params.model-version)-production
```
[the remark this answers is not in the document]
grdryn: Potentially, but actually I think it probably makes most sense to not have a default, so I'll take it out for now :+1:

test/e2e-tests/support/options.go
grdryn: Fixed, thanks

grdryn:
> it'd be great if you can also update the pipelines/README with the new parameters and how to use it

I've added a line to the "Other parameters section of that README now, thanks :+1:

grdryn:
> Passing the json to the make command is a bit weird though
TARGET_IMAGE_TAGS_JSON='[\"quay.io/jackdelahunt/edgetesting:latest\"]'

@jackdelahunt how would you feel if I dropped that parameter altogether, and just forced in the test to push to the internal registry like I mentioned in a comment? That would also mean the test setup wouldn't have to configure image registry credentials.

pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
devguyio: yeah I'm fine with the name

MarianMacik (commented):
Post-merge review, looks very good!

PR #212, opened 2024-02-20
Add tensorflow housing to go e2e-tests

jackdelahunt:
## Description
Adds the tensorflow housing pipeline to the go e2e-test suite

## Verify steps
- Log into a cluster with `oc`
- Set the list of enviroment variables that are needed for running the tests for both the `go-test` and `go-test-setup` make targets. All of the required fields are listed [here](https://github.com/jackdelahunt/ai-edge/blob/pipeline-testing/Makefile#L37-L44) and [here](https://github.com/jackdelahunt/ai-edge/blob/pipeline-testing/Makefile#L74-L76)
- Run `oc new-project test-namespace`
- Run `make AWS_SECRET=... AWS_ACCESS=...  ... go-test-setup`
- Run `make S3_BUCKET=...  ... go-test`
- All tests should pass

## How Has This Been Tested?

### DIFF
--- a/test/e2e-tests/tests/pipelines_test.go
+++ b/test/e2e-tests/tests/pipelines_test.go
@@ -11,8 +11,9 @@ import (
 )
 
 const (
-	AIEdgeE2EPipelineDirectoryRelativePath = "../../../pipelines/tekton/aiedge-e2e"
-	PipelineRunFileRelativePath            = AIEdgeE2EPipelineDirectoryRelativePath + "/aiedge-e2e.pipelinerun.yaml"
+	AIEdgeE2EPipelineDirectoryRelativePath       = "../../../pipelines/tekton/aiedge-e2e"
+	BikeRentalsPipelineRunFileRelativePath       = AIEdgeE2EPipelineDirectoryRelativePath + "/aiedge-e2e.bike-rentals.pipelinerun.yaml"
+	TensorflowHousingPipelineRunFileRelativePath = AIEdgeE2EPipelineDirectoryRelativePath + "/aiedge-e2e.tensorflow-housing.pipelinerun.yaml"
 )
 
 func CreateContext() context.Context {
@@ -79,7 +80,7 @@ func init() {
 	}
 }
 
-func Test_DefaultPipelineRun(t *testing.T) {
+func Test_PipelineRunsComplete(t *testing.T) {
 	ctx := CreateContext()
 
 	options, err := support.GetOptions()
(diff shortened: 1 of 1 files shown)


MarianMacik (approved):

jackdelahunt:
@MarianMacik this PR is a copy from an older one where me and @grdryn were still having a discussion about more changes to this. Just want to hold this until then

jackdelahunt:
I have a few questions about supporting more then one pipeline run and want to get feedback.

1. Should all of the configuration come from the test params from the make target? I think it makes it easier but means tests can be less varied
2. This pushes both pipeline runs and waits for them in a single test, maybe we want it one per test. I didn't do this because we would just be duplicating the test mostly. But maybe the answer to Q1 will change this 
3. The tensorflow housing pipeline run is getting the s3 bucket name set even though it uses git to fetch

jackdelahunt:
> or are you thinking of other things that could be configurable but aren't yet?
Not anything in particular, just that currently we enforce that all of the models come from the same bucket, or that we push to the same image repo. I guess it doesn't really matter as we are testing the pipelines themselves

> what pipeline scenario is it trying to verify? 
For me I was thinking each test is a pipeline run, but then when I did that it really just came down to changing which rile we read so it could just be done in a for loop so that is what I did. I think leaving it as is until we want to expand more is probably good. Not to start thinking about scaling when we just have two runs being tested 

> I'd guess that can be removed, my understanding is that it's being ignored unless the fetch-model parameter is changed to s3, right?
Well because they are all in the same test it gets set anyway but yes it doesn't have any affect in the end

grdryn:
> > or are you thinking of other things that could be configurable but aren't yet?
> 
> Not anything in particular, just that currently we enforce that all of the models come from the same bucket, or that we push to the same image repo. I guess it doesn't really matter as we are testing the pipelines themselves

For the image registry/repo, [this comment on my PR](https://github.com/opendatahub-io/ai-edge/pull/211/files#r1496302919) might have an alternative.

> 
> > what pipeline scenario is it trying to verify?
> 
> For me I was thinking each test is a pipeline run, but then when I did that it really just came down to changing which rile we read so it could just be done in a for loop so that is what I did. I think leaving it as is until we want to expand more is probably good. Not to start thinking about scaling when we just have two runs being tested

Yeah, we can always improve on this in future.

jackdelahunt:
> For the image registry/repo, [this comment on my PR](https://github.com/opendatahub-io/ai-edge/pull/211/files#r1496302919) might have an alternative.

That is interesting, where does the context come from there?

And if there is onthing more on this PR maybe it is ready for a LGTM. I think prow is failing but not because of my changes I don't think

grdryn (approved):
/lgtm

PR #214, opened 2024-02-21
RHOAIENG-3368|RHOAIENG-3461- Include Git model repo revision param and use it in the PR check

MarianMacik:
**JIRAs:**
* [RHOAIENG-3368](https://issues.redhat.com/browse/RHOAIENG-3368)
* [RHOAIENG-3461](https://issues.redhat.com/browse/RHOAIENG-3461)

## Description
This PR fixes 2 issues:
* Adds a parameter to be able to configure a Git revision to checkout when fetching the model from the Git repository.
* Sets this parameter to the [pull request ref](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally?platform=linux#modifying-an-inactive-pull-request-locally) when the PR check in OpenShift CI is run to ensure the latest changes are tested.

## How Has This Been Tested?
Locally on an OpenShift cluster.

## Merge criteria:

### DIFF
--- a/pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
+++ b/pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
@@ -15,16 +15,10 @@ spec:
     type: string
     description: "[S3 ONLY] S3 bucket name where the model is stored"
     default: ""
-  - name: containerfileGitServer
+  - name: git-containerfile-repo
     type: string
-    description: Root domain of the Git server where the Containerfile repository is located
-  - name: containerfileGitOrgName
-    type: string
-    description: Organization or Group name where the Containerfile repository is located
-  - name: containerfileGitRepoName
-    type: string
-    description: Name of the Git repository with the Containerfile
-  - name: containerfileGitRevision
+    description: Git repository where the Containerfile is stored
+  - name: git-containerfile-revision
     type: string
     description: Git revision to checkout when cloning the repository with the Containerfile
     default: "main"
@@ -163,9 +157,9 @@ spec:
   - name: git-clone-containerfile-repo
     params:
     - name: url
-      value: $(params.containerfileGitServer)/$(params.containerfileGitOrgName)/$(params.containerfileGitRepoName)
+      value: $(params.git-containerfile-repo)
     - name: revision
-      value: $(params.containerfileGitRevision)
+      value: $(params.git-containerfile-revision)
     - name: subdirectory
       value: /containerfile_repo/
     runAfter:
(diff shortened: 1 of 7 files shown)


pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
```diff
@@ -15,15 +15,19 @@ spec:
...
     type: string
-    description: Root domain of the Git server
-  - name: gitOrgName
+    description: Root domain of the Git server where the Containerfile repository is located
+  - name: containerfileGitOrgName
     type: string
-    description: Organization or Group name where the repo is located
-  - name: gitRepoName
+    description: Organization or Group name where the Containerfile repository is located
+  - name: containerfileGitRepoName
```
grdryn, line 24: This might be a good opportunity to turn these into a single parameter, similar to how `git-model-repo` is a single parameter, WDYT?

Are there any cases where we use these parts independently, or are we always constructing a single string composed of all 3 in the same order all the time?

MarianMacik: @grdryn actually we don't use them individually, I was thinking whether the whole or partial approach is better but it is quite common to input the URL as a whole, so I will change that. Thanks for input.

MarianMacik: Done.

grdryn (commented):
Thanks for the updated changes, they look good! :+1: 

Soft-lgtm to allow time for others to review before merging.

test/shell-pipeline-tests/common.sh
```diff
@@ -74,3 +74,11 @@ function createImageRegistrySecret() {
     sed -i "s|{{ IMAGE_REGISTRY_USERNAME }}|${USERNAME}|" "$IMAGE_REGISTRY_SECRET_PATH"
     sed -i "s|{{ IMAGE_REGISTRY_PASSWORD }}|${PASSWORD}|" "$IMAGE_REGISTRY_SECRET_PATH"
 }
+
+function usePRBranchInPipelineRunIfPRCheck() {
+  local -r PIPELINE_RUN_FILE_PATH=$1
+
+  if [ -n $PULL_NUMBER ]; then
+    sed -i "s|value: \"main\"|value: \"pull/$PULL_NUMBER/head\"|" "$PIPELINE_RUN_FILE_PATH"
```
LaVLaS, line 82: Do we need to specific about which parameter's  `value: main` we are targeting to prevent from replacing globally in the instance that the `git-containerfile-repo` and `git-model-repo` are different.  

Right now, we can defer that use case to setting the requirement that the model repo and containerfile repo are the same

MarianMacik: @LaVLaS This is just for the tests, so I think for now it is good as we want to have everything from the pull request branch for the PR check. The example shows main for both, so I just override both to the pull request branch. But yes, in real world these can be different revisions. Is that OK?

LaVLaS:
/lgtm

Another soft-lgtm to merge

PR #215, opened 2024-02-27
RHOAIENG-2758: Remove need to explicitly create imagestream

grdryn:
JIRA: https://issues.redhat.com/browse/RHOAIENG-2758

This is a preparatory change for RHOAIENG-2758, where we want to be able to use different image registries for the candidate image that the pipeline builds.

## Description

Before this change, we were explicitly creating the imagestream for the candidate image from the build task, just so that we could set the value of `spec.lookupPolicy.local` to `true`. While we could have left the imagestream to be implicitly created with default config just by pushing an image with the equivalent name, this value would not have had the desired value.

When the value of `spec.lookupPolicy.local` is at the default of false, then referring to an image by just `<name>:<tag>` without fully qualifying it with the registry domain, then it would default to the `docker.io` registry, and ignore the repository with that name in the internal registry in the cluster.

By setting the container image path to the fully qualified name, including the registry part (the same has how we specify it with in the parameter to the buildah task), then we don't need to care about the `spec.lookupPolicy.local` field, and can rely on the default creation of the ImageStream.

## How Has This Been Tested?

Pipeline has been run in a new namespace, where the ImageStream didn't exist.

## Merge criteria:

### DIFF
--- a/pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
+++ b/pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
@@ -188,18 +188,6 @@ spec:
     workspaces:
     - name: workspace
       workspace: build-workspace-pv
-  - name: create-imagestream
-    params:
-    - name: SCRIPT
-      value: oc create imagestream $(params.model-name)  --lookup-local=true --dry-run=client
-        -o yaml | oc apply -f -
-    - name: VERSION
-      value: latest
-    runAfter:
-    - check-model-and-containerfile-exists
-    taskRef:
-      kind: ClusterTask
-      name: openshift-client
   - name: build-container
     params:
     - name: IMAGE
@@ -221,7 +209,7 @@ spec:
     - name: SKIP_PUSH
       value: "false"
     runAfter:
-    - create-imagestream
+    - check-model-and-containerfile-exists
     taskRef:
       kind: ClusterTask
       name: buildah
@@ -254,7 +242,7 @@ spec:
                 app: $(params.model-name)-$(params.model-version)
             spec:
               containers:
-              - image: $(params.model-name):$(params.model-version)
+              - image: image-registry.openshift-image-registry.svc:5000/$(context.pipelineRun.namespace)/$(params.model-name):$(params.model-version)
                 name: $(params.model-name)-$(params.model-version)
                 livenessProbe:
                   failureThreshold: 10


pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
```diff
@@ -254,7 +242,7 @@ spec:
                 app: $(params.model-name)-$(params.model-version)
             spec:
               containers:
-              - image: $(params.model-name):$(params.model-version)
+              - image: image-registry.openshift-image-registry.svc:5000/$(context.pipelineRun.namespace)/$(params.model-name):$(params.model-version)
```
biswassri, line 245: Good simple and good to me, just curious what this change in image for? For the main task?

MarianMacik: The way I understand is to no limit ourselves to `<image>:<tag>` reference as this won't work with quay.io or other external image registries as it can work only in 2 situations:
* spec.lookupPolicy.local=true -> look for image stream in the namespace, which is created only in case an internal registry is used
* spec.lookupPolicy.local=false (default) -> look for docker.io image

which made me thinking @grdryn , don't we plan to use the image streams for external repositories? Because if we create an image stream all the time, with reference to the repository, then we could use the "short" reference and also the `retreive-build-image-info` task would not need to be changed:
https://github.com/opendatahub-io/ai-edge/blob/ad9123565759bb97711c56e2b40b1b13ac03f659/pipelines/tekton/aiedge-e2e/tasks/retrieve-build-image-info.task.yaml#L36

How will we get all these information without an image stream?

grdryn: @biswassri thanks for the review!

So this image field that we're commenting on is in the `Deployment` manifest that will be used in the testing of the candidate image that was built. This change doesn't actually change anything about the image or where it is. It's just that now that we're not explicitly creating an image stream with the local lookup policy enabled, we need to explicitly say which registry that it comes from.

The main task is to make it possible to use other registries than the internal registry in the OpenShift cluster for the candidate image. There are various reasons why that might be desired in different cases. So when the image isn't guaranteed to be in the internal registry, it doesn't make as much sense for us to use an ImageStream, and that's what this task is for: lessen the need for it.

Does that make sense? I _think_ that's what you were asking, but I'm not 100% sure I've answered the question that you've asked.

grdryn: > How will we get all these information without an image stream?

@MarianMacik That's what I'm working on figuring out at the moment. :) For the line that you linked, I believe it'll be something like `skopeo inspect --format "{{.Digest}}" <image tag ref>`, for example.

For this immediate PR, an ImageStream is still used (and the internal registry is still used), it's just that it's created implicitly, because we don't really need to create it explicitly.

MarianMacik: @grdryn But would image streams also work in case we import the image from external registry in an image stream? I guess this is also a supported configuration at least when looking at the [OCP docs](https://docs.openshift.com/container-platform/4.14/openshift_images/image-streams-manage.html#images-imagestream-external-image-tags_image-streams-managing).

What would be the benefit of not using image streams? Not having to update them when a new image is being pushed to an external repository?

grdryn: > But would image streams also work in case we import the image from external registry in an image stream? I guess this is also a supported configuration at least when looking at the [OCP docs](https://docs.openshift.com/container-platform/4.14/openshift_images/image-streams-manage.html#images-imagestream-external-image-tags_image-streams-managing).

I think we'd probably be able to create an ImageStream that represented either an internal or external image, yes. I don't think there's any benefit to it though, and it would just be extra complexity (I think we'd probably have to do things like manually sync the ImageStream after pushing the image to the external registry, so that we could use it later in the pipeline.

> What would be the benefit of not using image streams? Not having to update them when a new image is being pushed to an external repository?

Yes, this, and other things like in future if we want to enable testing on different environments than the local cluster where the image is built: the only place we really use the image stream (before this change) is in the test deployment, and if that was deployed to a different environment, then the imagestream wouldn't be accessible.

MarianMacik: @grdryn Makes sense, thanks.

MarianMacik (commented):
Looks good to me.

pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
biswassri: Makes sense to me too, thanks @grdryn

MarianMacik (approved):
LGTM :+1:

PR #216, opened 2024-02-27
RHOAIENG-437: Add instructions to fetch models from different s3 regions

Sara4994:
## Description

This PR adds instructions to the [credentials-s3.secret.template](https://github.com/opendatahub-io/ai-edge/blob/5960d4fc90c6f5aeb7c6a4eab5af5b1c2e972158/pipelines/tekton/aiedge-e2e/templates/credentials-s3.secret.yaml.template) about the optional endpoint_url and region params. [JIRA](https://issues.redhat.com/browse/RHOAIENG-437)

To fetch models from S3 in a different region from the IAM credentials, user can consider endpoint_url param as an optional one and pass the correct region to the credentials secret. This way region will take precedence over endpoint_url and allows the task to download model from the specified region.

## Merge criteria:

### DIFF
--- a/pipelines/tekton/aiedge-e2e/templates/credentials-s3.secret.yaml.template
+++ b/pipelines/tekton/aiedge-e2e/templates/credentials-s3.secret.yaml.template
@@ -7,5 +7,9 @@ metadata:
     app.kubernetes.io/part-of: rhoai-edge-pipelines
 
 stringData:
+  # The endpoint_url property is an optional one when using AWS S3, you can ignore it totally in case of AWS S3.
+  # Generally it takes priority over the region when specified. 
+  # If the endpoint_url is not specified it would default to an AWS endpoint based on region specified.
+  # Set the bucket region correctly.
   s3-storage-config: |+
     { "type": "s3", "access_key_id": "{{ AWS_ACCESS_KEY_ID }}", "secret_access_key": "{{ AWS_SECRET_ACCESS_KEY }}", "endpoint_url": "{{ S3_ENDPOINT }}", "region": "{{ S3_REGION }}" }


pipelines/tekton/aiedge-e2e/templates/credentials-s3.secret.yaml.template
```diff
@@ -7,5 +7,8 @@ metadata:
     app.kubernetes.io/part-of: rhoai-edge-pipelines
 
 stringData:
+  # The endpoint_url property is an optional one when using AWS S3, generally it takes priority over the region when specified. 
+  # If the endpoint_url is not specified it would default to an AWS endpoint based on region specified.
+  # Set the bucket region correctly.
```
MarianMacik, line 12: So the region is actually based on the credentials? And in case a user wants to use a bucket from a different region than credentials they would specify an endpoint_url from a different region?
In other words, `endpoint_url` is optional while `region` is mandatory?

MarianMacik: Thanks. How does it relate to the JIRA then? There it says `Currently the PipelineRuns will fail in the fetch-model-s3 step if the S3 bucket region is not the same region as the IAM credentials.` Based on what you said the S3 will intelligently handle this, if I specify a region us-west but bucket is in the us-east, it will be handled with an automatic redirect. So the S3 region doesn't have to be correct one but it has to be one from the regions provided by AWS. If that is so, I would reword the comment then.

grdryn: @MarianMacik I think the Jira issue wording was based on lack of complete understanding of the problem, and was likely based on always having the endpoint_url set. So if you had config where `region` was set to `eu-west-1` and corresponding endpoint_url of `https://s3.eu-west-1.amazonaws.com`, but the bucket was in `us-east-2`, then you'd get the behaviour described in the issue in Jira.

LaVLaS: >I think the Jira issue wording was based on lack of complete understanding of the problem, and was likely based on always having the endpoint_url set. 

Correct.  The original Jira was created based on my experience with boto3 connections to non-AWS S3 compatible endpoints where the endpoint_url AND [default] region where required for all client sessions.

Acceptance Criteria for this PR is that it should be clear to the user that `endpoint_url` and `region` are required for all non-AWS object storage servers BUT AWS S3 only requires the region

LaVLaS:
Adding this for historical reference, the use of the `s3.amazonaws.com` global endpoint is now classified a [Legacy Global Endpoint](https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html#deprecated-global-endpoint).

grdryn:
> Adding this for historical reference, the use of the `s3.amazonaws.com` global endpoint is now classified a [Legacy Global Endpoint](https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html#deprecated-global-endpoint).

That's an interesting page, I hadn't seen it, thanks. I don't think we should mention that endpoint in our docs, since it will get confusing. If I understand that page correctly, that legacy endpoint will only work for _some_ regions (those that existed before March 20th 2019).

Sara4994:
> > Adding this for historical reference, the use of the `s3.amazonaws.com` global endpoint is now classified a [Legacy Global Endpoint](https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html#deprecated-global-endpoint).
> 
> That's an interesting page, I hadn't seen it, thanks. I don't think we should mention that endpoint in our docs, since it will get confusing. If I understand that page correctly, that legacy endpoint will only work for _some_ regions (those that existed before March 20th 2019).

in that case this entire note wouldn't make sense? as we could just remove endpoint_url property from the template itself and just keep region alone?

grdryn:
> > > Adding this for historical reference, the use of the `s3.amazonaws.com` global endpoint is now classified a [Legacy Global Endpoint](https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html#deprecated-global-endpoint).
> > 
> > 
> > That's an interesting page, I hadn't seen it, thanks. I don't think we should mention that endpoint in our docs, since it will get confusing. If I understand that page correctly, that legacy endpoint will only work for _some_ regions (those that existed before March 20th 2019).
> 
> in that case this entire note wouldn't make sense? as we could just remove endpoint_url property from the template itself and just keep region alone?

For _Amazon AWS S3_, yes, we could remove the endpoint_url; but if someone wants to use some other implementation of the S3 API (of which there are many, such as if a user has a Minio instance deployed to the cluster, and that's what they're using for model storage), then the endpoint_url needs to be specified.

Sara4994:
> > > > Adding this for historical reference, the use of the `s3.amazonaws.com` global endpoint is now classified a [Legacy Global Endpoint](https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html#deprecated-global-endpoint).
> > > 
> > > 
> > > That's an interesting page, I hadn't seen it, thanks. I don't think we should mention that endpoint in our docs, since it will get confusing. If I understand that page correctly, that legacy endpoint will only work for _some_ regions (those that existed before March 20th 2019).
> > 
> > 
> > in that case this entire note wouldn't make sense? as we could just remove endpoint_url property from the template itself and just keep region alone?
> 
> For _Amazon AWS S3_, yes, we could remove the endpoint_url; but if someone wants to use some alternative service providing the S3 API (of which there are many, such as if a user has a Minio instance deployed to the cluster, and that's what they're using for model storage), then the endpoint_url needs to be specified.

Sara4994:
> Member

@grdryn cool, i have just tweaked the note a bit.

Sara4994:
> > > > Adding this for historical reference, the use of the `s3.amazonaws.com` global endpoint is now classified a [Legacy Global Endpoint](https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html#deprecated-global-endpoint).
> > > 
> > > 
> > > That's an interesting page, I hadn't seen it, thanks. I don't think we should mention that endpoint in our docs, since it will get confusing. If I understand that page correctly, that legacy endpoint will only work for _some_ regions (those that existed before March 20th 2019).
> > 
> > 
> > in that case this entire note wouldn't make sense? as we could just remove endpoint_url property from the template itself and just keep region alone?
> 
> For _Amazon AWS S3_, yes, we could remove the endpoint_url; but if someone wants to use some alternative service providing the S3 API (of which there are many, such as if a user has a Minio instance deployed to the cluster, and that's what they're using for model storage), then the endpoint_url needs to be specified.

@grdryn cool, i have just tweaked the note a bit.

grdryn:
Thanks @Sara4994, I'm fine with this now. I'll leave it for a while to see if anyone else wants to review. Otherwise I'll approve later.

MarianMacik (commented):
Looks good to me now.

LaVLaS:
/approve

Adding approval also based on the additional LGTMs above

PR #217, opened 2024-02-27
RHOAIENG-1119: Don't rely on imagestream in image info task

grdryn:
This PR stacks upon #215 as another preparatory commit for [RHOAIENG-1119](https://issues.redhat.com//browse/RHOAIENG-1119). It will be held until that one is merged, but it can still be reviewed, by just reviewing the new commit(s).

## Description
With this change, we now get info for the image from the image metadata using `skopeo inspect`, rather than from the ImageStream resource in OpenShift.

## How Has This Been Tested?

- `make go-test` runs successfully
- Example pipelineruns executed manually with success

## Merge criteria:

### DIFF
--- a/pipelines/tekton/aiedge-e2e/tasks/retrieve-build-image-info.task.yaml
+++ b/pipelines/tekton/aiedge-e2e/tasks/retrieve-build-image-info.task.yaml
@@ -21,7 +21,7 @@ spec:
   - name: images-url
   steps:
   - name: get-image-sha
-    image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
+    image: registry.access.redhat.com/ubi9/skopeo
     args:
       - "$(params.target-image-tag-references[*])"
     script: |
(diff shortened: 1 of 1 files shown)


pipelines/tekton/aiedge-e2e/tasks/retrieve-build-image-info.task.yaml
```diff
@@ -33,18 +33,18 @@ spec:
...
+      export DOCKER_IMAGE_REF=$(skopeo inspect --format '{{.Name}}@{{.Digest}}' docker://image-registry.openshift-image-registry.svc:5000/$(params.namespace)/$(params.model-name):$(params.model-version)) ;
       if [[ $DOCKER_IMAGE_REF != *"$(params.buildah-sha)"* ]]; then
-        echo "ImageStreamTag doesn't contain the correct image SHA"
+        echo "Candidate image tag does not contain the correct manifest digest after push"
         exit 1 ;
       fi
       echo -n $DOCKER_IMAGE_REF | tee $(results.internal-image-url.path) ;
       echo ;
-      oc get -n $(params.namespace) -o jsonpath='{.image.dockerImageMetadata.Size}' imagestreamtag/$(params.model-name):$(params.model-version) | tee $(results.internal-image-size.path) ;
+      echo $(($(skopeo inspect --format '{{range .LayersData}}+{{.Size}}{{end}}' docker://$DOCKER_IMAGE_REF))) | tee $(results.internal-image-size.path) ;
```
MarianMacik, line 43: Just FYI, when I tried this, when inspecting `docker://` image, it returns uncompressed size, when using `docker-daemon:sha256:<sha256-of-local-docker-registry>` it seems to return the full size.
E.g.:
`skopeo inspect docker://docker.io/library/fedora@sha256:830a6bcb5c9b30bfa77479b747a904bed02151e9de7ceeb8a1d117b574571fa4` (fedora:latest)
returns 
``` json
"Size": 64614010
```
whereas
`skopeo inspect docker-daemon:sha256:5f878605acd6f24f3f11e4085544caf61b046378b31d6b3aa0677a7a3dbe0a17`
which is my local IMAGE ID for the same image returns
```json
"Size": 182162432
```
and docker inspect for that same image returns
```json
"Size": 176658692
```
The last 2 outputs are from local image, the first one is for a remote, compressed image.

I think for now it's OK. Just in case we need to have constraints about image size we should be more careful what we are measuring.

grdryn: Yeah, that's interesting. I was actually initially just going to omit this bit of information (and maybe one or two others also), since I'm not sure why we want it or how we'll use it.

It does appear to be the compressed size that we were getting previously: I imported an image into an imagestream & compared. E.g.:

```
# import image to imagestream
oc import-image fedora --from=quay.io/fedora/fedora --all --confirm

# then check the size of a tag in the imagestream
oc get -o jsonpath='{.image.dockerImageMetadata.Size}' imagestreamtag/fedora:39
68672016

# compare that to the new approach of adding layer sizes in original repo
echo $(($(skopeo inspect --format '{{range .LayersData}}+{{.Size}}{{end}}' docker://quay.io/fedora/fedora:39)))
68670700
```

^there's a slight discrepancy (~1KiB), which I'd guess is the manifest referring to the layers?

MarianMacik: This metric is for the future when we are planning to check the real image size and have a test for it, if the image is above the limit, the pipeline would likely fail or something similar.

Also for my other images with more than one layer the numbers are slightly off, roughly by 5 MB. Interestingly skopeo reports bigger sum than docker inspect. But it may be also due to rounding or different way of calculation by skopeo.

pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
```diff
@@ -254,7 +242,7 @@ spec:
                 app: $(params.model-name)-$(params.model-version)
             spec:
               containers:
-              - image: $(params.model-name):$(params.model-version)
+              - image: image-registry.openshift-image-registry.svc:5000/$(context.pipelineRun.namespace)/$(params.model-name):$(params.model-version)
```
StevenTobin, line 245: Should this be configurable? Would we expect this to change.

grdryn: Yes, it will become configurable as part of #218. This PR/commit is just preparation for that

grdryn:
Going to close and re-open in an attempt to get GitHub to automatically update the PR view (it's out of date and showing a commit that's already on main).

grdryn:
> Going to close and re-open in an attempt to get GitHub to automatically update the PR view (it's out of date and showing a commit that's already on main).

Wow, that actually worked. This commit should now be even easier to review

Sara4994:
This /lgtm when test goes green

grdryn:
@Sara4994 thanks for the review! You don't need to wait for the test status check to go green before adding the /lgtm (or approve using the review feature), since it won't automatically merge until it has both the lgtm label and the test job passes. So please just /lgtm :slightly_smiling_face:

biswassri (commented):
/lgtm

Sara4994 (approved):

PR #218, opened 2024-02-28
RHOAIENG-1119: Add new candidate-image-tag-reference parameter

grdryn:
This PR stacks upon #215 and #217. It will be held until those are merged, but it can still be reviewed, by just reviewing the new commit(s). This PR is the main change for the [RHOAIENG-1119](https://issues.redhat.com//browse/RHOAIENG-1119).

## Description
This will allow overriding the internal image registry reference, and
use any other image registry, repo, and tag instead.

There's a default set for the new param to maintain the old behaviour,
and the same value is set in the example PipelineRun files by default.

## How Has This Been Tested?

- `make go-test` run successfully
- Example pipelines run manually using default value for new parameter, proving that existing behaviour works as expected.
- Example pipelines run, with the new parameter set to `quay.io/grdryn/$(params.model-name):$(params.model-version)-candidate`, proving that an external registry can easily be used for the candidate image, as can be seen in the following screenshot (see tag `1-candidate`):
image

## Merge criteria:

### DIFF
--- a/pipelines/tekton/aiedge-e2e/aiedge-e2e.bike-rentals.pipelinerun.yaml
+++ b/pipelines/tekton/aiedge-e2e/aiedge-e2e.bike-rentals.pipelinerun.yaml
@@ -30,6 +30,8 @@ spec:
     value: "main"
   - name: test-endpoint
     value: "invocations"
+  - name: candidate-image-tag-reference
+    value: image-registry.openshift-image-registry.svc:5000/$(context.pipelineRun.namespace)/$(params.model-name):$(params.model-version)-candidate
   - name: target-image-tag-references
     value:
     - quay.io/rhoai-edge/$(params.model-name):$(params.model-version)-$(context.pipelineRun.uid)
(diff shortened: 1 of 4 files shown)


MarianMacik (commented):
Looks good to me. I will leave it for others to have a look as well.

biswassri:
Looks good to me too. Should I go ahead and add a `/lgtm`?

grdryn:
@biswassri you're absolutely welcome to add /lgtm here too :slightly_smiling_face: It won't get merged until either the `ci/prow/test-ai-edge` check passes, or we explicitly skip it.

> Looks good to me too. @grdryn just a quick question, the candidate tag in the bike-rental pipeline run is just for an example right?

Yes, both of our PipelineRun files are just examples, so the value for the candidate tag that I've added to them (they both have the same value) is just adding to the examples. You are welcome to change it to any other sane value in any other registry/repo/tag, as long as the pipeline service account has access to be able to push there (done by linking a secret, typically).

PR #220, opened 2024-02-28
RHOAIENG-3461 - Fix

MarianMacik:
Fix for #214.

Forgot to include the quotes, my bad. Without them, there is only 1 argument (`-n`) and the condition is [always true](https://www.gnu.org/software/bash/manual/bash.html#index-test):
```
1 argument
    The expression is true if, and only if, the argument is not null.
```

## Merge criteria:

### DIFF
--- a/test/shell-pipeline-tests/common.sh
+++ b/test/shell-pipeline-tests/common.sh
@@ -78,7 +78,7 @@ function createImageRegistrySecret() {
 function usePRBranchInPipelineRunIfPRCheck() {
   local -r PIPELINE_RUN_FILE_PATH=$1
 
-  if [ -n $PULL_NUMBER ]; then
+  if [ -n "$PULL_NUMBER" ]; then
     sed -i "s|value: \"main\"|value: \"pull/$PULL_NUMBER/head\"|" "$PIPELINE_RUN_FILE_PATH"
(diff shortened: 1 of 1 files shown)


test/shell-pipeline-tests/common.sh
```diff
@@ -78,7 +78,7 @@ function createImageRegistrySecret() {
 function usePRBranchInPipelineRunIfPRCheck() {
   local -r PIPELINE_RUN_FILE_PATH=$1
 
-  if [ -n $PULL_NUMBER ]; then
+  if [ -n "$PULL_NUMBER" ]; then
```
grdryn, line 81: One thing to consider is to have `shellcheck` for these shell scripts. Here's what it would have printed for this specific line:

```
In test/shell-pipeline-tests/common.sh line 81:
  if [ -n $PULL_NUMBER ]; then
          ^----------^ SC2070 (error): -n doesn't work with unquoted arguments. Quote or use [[ ]].
          ^----------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
  if [ -n "$PULL_NUMBER" ]; then
```

grdryn (approved):
/lgtm :+1:

PR #222, opened 2024-03-05
RHOAIENG-3457: Add an inital root CLI

devguyio:
## Description

This PR is an initial PR for adding a basic CLI that just lists the namespaces as a validation to the functionality. Next PRs will remove that and add the first command.

It also adds make targets to build and run the CLI and run any CLI tests.

## How Has This Been Tested?

Run the following make command
```shell
make cli-run
```

## Merge criteria:


pkg/commands/root_test.go
```diff
@@ -0,0 +1,65 @@
...
+			ns2 := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "test-namespace-2"}}
+			clientSet.CoreV1().Namespaces().Create(context.TODO(), ns1, metav1.CreateOptions{})
+			clientSet.CoreV1().Namespaces().Create(context.TODO(), ns2, metav1.CreateOptions{})
+		},
+	)
+
+	// Create a pipe to capture the output
+	r, w, _ := os.Pipe()
+	// Save the original stdout
+	originalOut := os.Stdout
```
MarianMacik, line 54: I think it would be better to replace this with:
```go
rootCmd.SetOut(buf) // no need for a pipe
```
which will redirect output of just the `rootCmd.Run()` call. This requires using `cmd.Printf("Namespace: %s\n", ns.Name)` instead of the `fmt` version in the root.go.

Right now any output like `fmt.Printf()` which is done after the `os.Stdout = w` will not show up on the terminal, so it took me a while to figure out why the tests were exiting with status 1. The issue was missing `KUBECONFIG`, but the information was not displayed as it was output to the pipe and the whole program exited before the pipe was copied to buffer. If you use cobra's printing, then only the output of the command itself will be put to the buffer. Everything from the tests or init commands will be output to the normal Stdout. This makes the code more testable I believe. True, you cannot redirect the output of that init function printing `KUBECONFIG not set.` but the program would exit anyway just after printing this, so there won't be any possible test for it anyway (other than running the CLI as an external process).

pkg/commands/root.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// initConfig reads in config file and ENV variables if set.
+func initConfig() {
+	viper.AutomaticEnv() // read in environment variables that match
+
+	if kubeconfig = viper.GetString("KUBECONFIG"); kubeconfig == "" {
+		fmt.Printf("KUBECONFIG not set.")
+		os.Exit(1)
+	}
```
MarianMacik, line 76: Can we maybe default to `$HOME/.kube/config`? Or maybe document this as a prerequisite.

Makefile
```diff
@@ -90,3 +90,12 @@ endif
 test:
 	@(./test/shell-pipeline-tests/seldon-bike-rentals/pipelines-test-seldon-bike-rentals.sh)
 	@(./test/shell-pipeline-tests/openvino-tensorflow-housing/pipelines-test-openvino-tensorflow-housing.sh)
+
+cli-test:
+	${GO} test ./cmd/... ./... -v
```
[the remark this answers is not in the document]
devguyio: Done

cobra.yaml
```diff
@@ -0,0 +1,4 @@
+author: Open Data Hub Authors
+license: apache
+useViper: true
+
```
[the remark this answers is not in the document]
devguyio: Done

pkg/commands/root_test.go
devguyio: Thanks for the nice suggestions! Done.

pkg/commands/root.go
devguyio: Done

devguyio: done

MarianMacik (commented):
Looks good to me now!

Sara4994:
@devguyio can we include some readme instructions on how one can install and run cli. May be basic ones to start with(same instructions we had in the gitlab repo)

devguyio:
Holding this in case we decided to go with #226 instead
/hold

cli/pkg/commands/root_test.go
```diff
@@ -0,0 +1,52 @@
+/*
+Copyright 2023 KStreamer Authors
```
grdryn, line 2: Is this copyright assignment correct, or should it be 2024 Open Data Hub Authors, like the other files?

PR #229, opened 2024-03-17
RHOAIENG-2787: Let ArgoCD create namespace, rather than using ACM Policy

grdryn:
JIRA: https://issues.redhat.com/browse/RHOAIENG-2787

This is a preparatory commit for [RHOAIENG-2787](https://issues.redhat.com//browse/RHOAIENG-2787), as not needing to use ACM Policy will make things a little easier for non-ACM cases.

## Description

This works as expected as long as the namespace is specified in the GitOps repo, and has the correct `argocd.argoproj.io/managed-by` label value.
We already had the namespace specified in the GitOps repo, but we didn't include the label.

More information on this label can be found here:
https://docs.openshift.com/gitops/1.11/argocd_instance/setting-up-argocd-instance.html

## How Has This Been Tested?

- Created both example applications through ACM successfully, and noticed that they were created as expected on the edge clusters
- Deleted both of the apps successfully on the ACM hub cluster, and noticed that all of the resources, including the namespaces, were deleted on the edge clusters.

## Merge criteria:

### DIFF
--- a/acm/odh-edge/base/namespace.yaml
+++ b/acm/odh-edge/base/namespace.yaml
@@ -3,3 +3,9 @@ kind: Namespace
 metadata:
   # edit this to the namespace where the model is deployed
   name: model-to-edge
+  labels:
+    # If your OpenShift GitOps (or ArgoCD) is in a different namespace
+    # than openshift-gitops, then update this label value to match.
+    # More information:
+    # https://docs.openshift.com/gitops/1.11/argocd_instance/setting-up-argocd-instance.html
+    argocd.argoproj.io/managed-by: openshift-gitops
--- a/acm/registration/near-edge/overlays/bike-rental-app/kustomization.yaml
+++ b/acm/registration/near-edge/overlays/bike-rental-app/kustomization.yaml
@@ -24,14 +24,6 @@ patches:
     group: argoproj.io
     version: v1alpha1
     kind: ApplicationSet
-- patch: |-
-    - op: replace
-      path: /spec/policy-templates/0/objectDefinition/metadata/name
-      value: bike-rental-app-ns-has-argo-label
-  target:
-    group: policy.open-cluster-management.io
-    version: v1
-    kind: Policy
 
 replacements:
 - source:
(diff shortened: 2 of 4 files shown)


acm/odh-edge/base/namespace.yaml
```diff
@@ -3,3 +3,5 @@ kind: Namespace
 metadata:
   # edit this to the namespace where the model is deployed
   name: model-to-edge
+  labels:
+    argocd.argoproj.io/managed-by: openshift-gitops
```
grdryn, line 7: @LaVLaS this is the label that I was saying would need to be updated to have a different value if trying to use it with upstream ArgoCD. I'd have thought it would be the ArgoCD instance name that would be the value, but this page seems to indicate that it's the _namespace_ that ArgoCD is in (I'm not sure if that's correct): https://docs.openshift.com/gitops/1.11/argocd_instance/setting-up-argocd-instance.html

In any case, I think probably in the upstream case, the ArgoCD instance and namespace are probably both called `argocd`, and with OpenShift GitOps, they're both called `openshift-gitops`.

LaVLaS: Ok. So for right now, we just need to be aware that this needs to change if someone is using upstream ArgoCD?

If so, can we add comment just as a reminder?  Otherwise it's LGTM

grdryn: > Ok. So for right now, we just need to be aware that this needs to change if someone is using upstream ArgoCD?

Yes :+1: 

> If so, can we add comment just as a reminder? Otherwise it's LGTM

Updated, thanks!

acm/registration/near-edge/overlays/bike-rental-app/kustomization.yaml
```diff
@@ -54,23 +46,3 @@ replacements:
       kind: PlacementBinding
```
MarianMacik, line 46: This can be removed too.

acm/registration/near-edge/overlays/tensorflow-housing-app/kustomization.yaml
```diff
@@ -54,23 +46,3 @@ replacements:
       kind: PlacementBinding
```
MarianMacik, line 46: This as well.

MarianMacik (commented):
Looks good (forgot to approve), I will open a PR to fix those unneeded replacements I noticed only now.

PR #230, opened 2024-03-19
RHOAIENG-4166: Add support for self signed certs in the MLOps pipeline

jackdelahunt:
## Description
What this does:
- A self signed cert can be passed as a parameter to the MLops pipeline and can be used for use with private services throughout the pipeline, but for ones that only use git
- Documents the process and an optional step in the pipeline docs

## How Has This Been Tested?
- Use the e2e test provided in the repo
- Update the pipeline run to use a git repo that uses SSC (ping me for a repo and a cert to use!)
- Get access to that self signed cert and use it's path in the e2e tests as below
- Run 
```bash
make AWS_SECRET_ACCESS_KEY=... AWS_ACCESS_KEY_ID=... S3_REGION=... S3_ENDPOINT=... IMAGE_REGISTRY_USERNAME=...  IMAGE_REGISTRY_PASSWORD=... SELF_SIGNED_CERT=<PATH_TO_SSC> S3_BUCKET=... NAMESPACE=... TARGET_IMAGE_TAGS_JSON=... go-test-setup
```

- And then run 
```bash
make AWS_SECRET_ACCESS_KEY=... AWS_ACCESS_KEY_ID=... S3_REGION=... S3_ENDPOINT=... IMAGE_REGISTRY_USERNAME=...  IMAGE_REGISTRY_PASSWORD=... SELF_SIGNED_CERT=<PATH_TO_SSC> S3_BUCKET=... NAMESPACE=... TARGET_IMAGE_TAGS_JSON=... go-test
```

## Merge criteria:

- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work

PR #241, opened 2024-04-03
RHOAIENG-4873: Kserve self signed certs 

jackdelahunt:
## Description
Add support for using self signed certs with kserve to download models from an object store 

## How Has This Been Tested?
- Use the e2e test provided in the repo
- Setup e2e tests as normal
```bash
make AWS_SECRET_ACCESS_KEY=... AWS_ACCESS_KEY_ID=... S3_REGION=... S3_ENDPOINT=... IMAGE_REGISTRY_USERNAME=...  IMAGE_REGISTRY_PASSWORD=... S3_BUCKET=... NAMESPACE=... TARGET_IMAGE_TAGS_JSON=... go-test-setup
```
- Set the new env variable `S3_SELF_SIGNED_CERT` as the path to the cert

- And then run 
```bash
make AWS_SECRET_ACCESS_KEY=... AWS_ACCESS_KEY_ID=... S3_REGION=... S3_ENDPOINT=... IMAGE_REGISTRY_USERNAME=...  IMAGE_REGISTRY_PASSWORD=...  S3_BUCKET=... NAMESPACE=... TARGET_IMAGE_TAGS_JSON=... S3_SELF_SIGNED_CERT=... go-test
```

## Merge criteria:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,13 +34,14 @@ GO=go
 GOFLAGS=""
 
 # requires:
-#	AWS_SECRET_ACCESS_KEY   - Secret from AWS
-#	AWS_ACCESS_KEY_ID       - Access key from AWS
-#	S3_REGION               - Region of the bucket used to store the model
-#	S3_ENDPOINT             - Endpint of the bucket
-#	IMAGE_REGISTRY_USERNAME - quay.io username
-#	IMAGE_REGISTRY_PASSWORD - quay.io password
-#	SELF_SIGNED_CERT        -  Local path to the self  signed cert to be used in pipeline (optional)
+#	AWS_SECRET_ACCESS_KEY		- Secret from AWS
+#	AWS_ACCESS_KEY_ID		- Access key from AWS
+#	S3_REGION			- Region of the bucket used to store the model
+#	S3_ENDPOINT			- Endpint of the bucket
+#	IMAGE_REGISTRY_USERNAME		- quay.io username
+#	IMAGE_REGISTRY_PASSWORD		- quay.io password
+#	GIT_SELF_SIGNED_CERT		- Self signed cert to be used in pipeline by git (optional)
+#	S3_SELF_SIGNED_CERT		- Self signed cert to be used in pipeline by kserve (optional)
 go-test-setup:
 ifndef AWS_SECRET_ACCESS_KEY
 	$(error AWS_SECRET_ACCESS_KEY is undefined)
@@ -60,18 +61,22 @@ endif
 ifndef IMAGE_REGISTRY_PASSWORD
 	$(error IMAGE_REGISTRY_PASSWORD is undefined)
 endif
-ifdef SELF_SIGNED_CERT
-	oc create configmap self-signed-cert --from-file=ca-bundle.crt=${SELF_SIGNED_CERT}
+ifdef GIT_SELF_SIGNED_CERT
+	oc create configmap git-self-signed-cert --from-file=ca-bundle.crt=${GIT_SELF_SIGNED_CERT}
+endif
+ifdef S3_SELF_SIGNED_CERT
+	oc create configmap s3-self-signed-cert --from-file=ca-bundle.crt=${S3_SELF_SIGNED_CERT}
 endif
 	@sed -e "s#{{ AWS_SECRET_ACCESS_KEY }}#${AWS_SECRET_ACCESS_KEY}#g" -e "s#{{ AWS_ACCESS_KEY_ID }}#${AWS_ACCESS_KEY_ID}#g" -e "s#{{ S3_REGION }}#${S3_REGION}#g" -e "s#{{ S3_ENDPOINT }}#${S3_ENDPOINT}#g" pipelines/tekton/aiedge-e2e/templates/credentials-s3.secret.yaml.template | oc create -f -
 	@sed -e "s#{{ IMAGE_REGISTRY_USERNAME }}#${IMAGE_REGISTRY_USERNAME}#g" -e "s#{{ IMAGE_REGISTRY_PASSWORD }}#${IMAGE_REGISTRY_PASSWORD}#g" pipelines/tekton/aiedge-e2e/templates/credentials-image-registry.secret.yaml.template | oc create -f -
 	oc secret link pipeline credentials-image-registry
 
 # requires:
-#	S3_BUCKET               - Name of S3 bucket that contains the models
-#	NAMESPACE               - Cluster namespace that tests are run in
-#	TARGET_IMAGE_TAGS_JSON  - JSON array of image tags that the final image will be pushed to. E.g. '["quay.io/user/model-name:e2e-test"]'
-#	SELF_SIGNED_CERT        - Local path to the self signed cert to be used in pipeline (optional)
+#	S3_BUCKET	- Name of S3 bucket that contains the models
+#	NAMESPACE	- Cluster namespace that tests are run in
+#	TARGET_IMAGE_TAGS_JSON	- JSON array of image tags that the final image will be pushed to. E.g. '["quay.io/user/model-name:e2e-test"]'
+#	GIT_SELF_SIGNED_CERT	- Self signed cert to be used in pipeline by git (optional)
+#	S3_SELF_SIGNED_CERT	- Self signed cert to be used in pipeline when fetching models from S3 (optional)
 go-test:
 ifndef S3_BUCKET
 	$(error S3_BUCKET is undefined)
@@ -82,7 +87,7 @@ endif
 ifndef TARGET_IMAGE_TAGS_JSON
 	$(error TARGET_IMAGE_TAGS_JSON is undefined)
 endif
-	(cd test/e2e-tests/tests && S3_BUCKET=${S3_BUCKET} TARGET_IMAGE_TAGS_JSON=${TARGET_IMAGE_TAGS_JSON} NAMESPACE=${NAMESPACE} SELF_SIGNED_CERT=${SELF_SIGNED_CERT} ${GO} test -timeout 30m)
+	(cd test/e2e-tests/tests && S3_BUCKET=${S3_BUCKET} TARGET_IMAGE_TAGS_JSON=${TARGET_IMAGE_TAGS_JSON} NAMESPACE=${NAMESPACE} GIT_SELF_SIGNED_CERT=${GIT_SELF_SIGNED_CERT} S3_SELF_SIGNED_CERT=${S3_SELF_SIGNED_CERT} ${GO} test -timeout 30m)
 
 test:
 	@(./test/shell-pipeline-tests/seldon-bike-rentals/pipelines-test-seldon-bike-rentals.sh)
--- a/Makefile
+++ b/Makefile
@@ -34,14 +34,14 @@ GO=go
 GOFLAGS=""
 
 # requires:
-#	AWS_SECRET_ACCESS_KEY		- Secret from AWS
-#	AWS_ACCESS_KEY_ID		- Access key from AWS
-#	S3_REGION			- Region of the bucket used to store the model
-#	S3_ENDPOINT			- Endpint of the bucket
-#	IMAGE_REGISTRY_USERNAME		- quay.io username
-#	IMAGE_REGISTRY_PASSWORD		- quay.io password
-#	GIT_SELF_SIGNED_CERT		- Self signed cert to be used in pipeline by git (optional)
-#	S3_SELF_SIGNED_CERT		- Self signed cert to be used in pipeline by kserve (optional)
+#   AWS_SECRET_ACCESS_KEY       - Secret from AWS
+#   AWS_ACCESS_KEY_ID           - Access key from AWS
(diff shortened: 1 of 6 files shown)


grdryn:
@jackdelahunt I know this is marked as Draft, but just wondering if it's actually needed? What's the overlap with #230?

jackdelahunt:
@grdryn sorry I didn't update this much, this is for self signed certs with kserve. The other PR is just for git

pipelines/tekton/aiedge-e2e/tasks/kserve-download-model.yaml
```diff
@@ -17,13 +17,16 @@ spec:
...
   - name: download-model-s3
-    image: quay.io/opendatahub/kserve-storage-initializer:v0.11
+    image: quay.io/opendatahub/kserve-storage-initializer:v0.11.1.3 
     script: |
       mkdir -p $(workspaces.workspace.path)/model_dir-$(params.model-name)/
 
       export S3_URL="s3://$(params.s3-bucket-name)/$(params.model-name)"
       echo -n $S3_URL | tee $(results.s3-url.path) ;
 
+      export CA_BUNDLE_CONFIGMAP_NAME=ssl-ca-directory
```
LaVLaS, line 27: Does this actually query for a ConfigMap named `ssl-ca-directory`?  If so, is there a kserve documentation about what is expected for the ConfigMap?

LaVLaS: To answer my own question: From [kserver v11.1.3/storage.py](https://github.com/opendatahub-io/kserve/blob/v0.11.1.3/python/kserve/kserve/storage/storage.py#L183), it looks like non-null `CA_BUNDLE_CONFIGMAP_NAME` is the logic check for getting the custom cert path in `AWS_CA_BUNDLE` and in this storage-initializer context there is no k8s client reading an actual ConfigMap

pipelines/tekton/aiedge-e2e/tasks/kserve-download-model.yaml
```diff
@@ -17,13 +17,16 @@ spec:
...
+      export AWS_CA_BUNDLE=$(workspaces.ssl-ca-directory.path)/cabundle.crt
```
LaVLaS, line 28: We are using `cabundle.crt` here but the git-clone task defaults to `ca-bundle.crt`  We should have some parity between the file name between this task and the git-clone task assuming that the value of `AWS_CA_BUNDLE` is the actual CA bundle filename we are mounting to the file

pipelines/tekton/aiedge-e2e/tasks/kserve-download-model.yaml
```diff
@@ -32,3 +35,7 @@ spec:
     name: workspace
   - description: The S3 secret.
     name: s3-secret
+  - description: |
```
LaVLaS, line 38: We'll have to set this as an `optional: true` workspace to support the workflows where a custom cert is not required for the Pipeline to run successfully and update the script to populate the env vars only when this workspace is mounted and the appropriate file exists

pipelines/tekton/aiedge-e2e/tasks/kserve-download-model.yaml
jackdelahunt: Yes, this was a temperaory change because depending on how you use kserve you need that exact name but these have now been seperated out to s3 and git and the names are the same as it doens't affect kserve

pipelines/tekton/aiedge-e2e/tasks/kserve-download-model.yaml
jackdelahunt: Yes, these were just while it was a draft and I was testing. Using self signed certs is now optional for both git and s3

pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
```diff
@@ -131,7 +131,7 @@ spec:
     - name: output
       workspace: build-workspace-pv
     - name: ssl-ca-directory
-      workspace: ssl-certs
+      workspace: git-ssl-certs
```
grdryn, line 134:
```suggestion
      workspace: git-tls-certs
```

Also, should it be `-certs` or `-cert` now?

grdryn (commented):
- I think we should prefer the term TLS cert rather than SSL cert.
- Where we say "self-signed cert", should we instead be saying something like "certificate authority"? Just because a cert isn't trusted by default, doesn't mean that it's "self-signed", does it? It could be signed by a centralized CA owned by the users organisation, and I guess that's what they'd need to import here in that case, is it?

pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
LaVLaS: We should keep this as `git-ssl-*` to maintain clarity between dependencies (boto3, aws s3, git, ...) that still publicly reference ssl within their respective tools (`use_ssl`, `--no-verify-ssl`, `http.sslVerify`, ...)

pipelines/tekton/aiedge-e2e/tasks/kserve-download-model.yaml
```diff
@@ -32,3 +37,8 @@ spec:
...
+      A workspace containing CA certificates, this will be used by kserve to
```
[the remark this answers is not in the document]
jackdelahunt: 👍

test/e2e-tests/README.md
```diff
@@ -21,6 +21,9 @@ The following enviroment variables are optional. You may still need to set them
 
 - `SELF_SIGNED_CERT` - A path on your local machine to a self signed cert to be used in pipeline
```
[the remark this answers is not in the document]
jackdelahunt: Yep it should of been removed

Makefile
```diff
@@ -34,13 +34,14 @@ GO=go
...
-#	IMAGE_REGISTRY_PASSWORD - quay.io password
-#	SELF_SIGNED_CERT        -  Local path to the self  signed cert to be used in pipeline (optional)
+#	AWS_SECRET_ACCESS_KEY		- Secret from AWS
+#	AWS_ACCESS_KEY_ID		- Access key from AWS
+#	S3_REGION			- Region of the bucket used to store the model
+#	S3_ENDPOINT			- Endpint of the bucket
+#	IMAGE_REGISTRY_USERNAME		- quay.io username
+#	IMAGE_REGISTRY_PASSWORD		- quay.io password
+#	GIT_SELF_SIGNED_CERT		- Self signed cert to be used in pipeline by git (optional)
+#	S3_SELF_SIGNED_CERT		- Self signed cert to be used in pipeline by kserve (optional)
```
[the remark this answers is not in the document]
jackdelahunt: 👍

Makefile
```diff
@@ -60,18 +61,22 @@ endif
...
 # requires:
-#	S3_BUCKET               - Name of S3 bucket that contains the models
-#	NAMESPACE               - Cluster namespace that tests are run in
-#	TARGET_IMAGE_TAGS_JSON  - JSON array of image tags that the final image will be pushed to. E.g. '["quay.io/user/model-name:e2e-test"]'
-#	SELF_SIGNED_CERT        - Local path to the self signed cert to be used in pipeline (optional)
+#	S3_BUCKET	- Name of S3 bucket that contains the models
+#	NAMESPACE	- Cluster namespace that tests are run in
+#	TARGET_IMAGE_TAGS_JSON	- JSON array of image tags that the final image will be pushed to. E.g. '["quay.io/user/model-name:e2e-test"]'
+#	GIT_SELF_SIGNED_CERT	- Self signed cert to be used in pipeline by git (optional)
+#	S3_SELF_SIGNED_CERT	- Self signed cert to be used in pipeline when fetching models from S3 (optional)
```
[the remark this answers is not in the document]
jackdelahunt: 👍

test/e2e-tests/README.md
```diff
@@ -21,6 +21,9 @@ The following enviroment variables are optional. You may still need to set them
...
 
+- `GIT_SELF_SIGNED_CERT` - Self signed cert to be used by git when cloning (optional)
+- `S3_SELF_SIGNED_CERT` - Self signed cert to be used for S3 when pulling models (optional)
```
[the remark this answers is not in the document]
jackdelahunt: 👍

pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
jackdelahunt: Yes SSL is still used even though I agree TLS is better. I changed to not be plural though 👍

jackdelahunt:
> * Where we say "self-signed cert", should we instead be saying something like "certificate authority"? 

So the configmaps that get mounted to the pipeline runs are called `git-self-signed-cert` and `s3-self-signed-cert`. What should these be renamed to? Just `git-cert`, `git-CA-cert` or `git-signed-cert`?

LaVLaS:
> * I think we should prefer the term TLS cert rather than SSL cert.

While I agree TLS is the correct term there is potential confusion with underlying dependencies that still use the term `SSL` for their public confgs & apis (See https://github.com/opendatahub-io/ai-edge/pull/241#discussion_r1562663813
).  In documentation, we could enforce the use of `SSL/TLS` where both apply or just `tls` based features/configs that we own completely

> Where we say "self-signed cert", should we instead be saying something like "certificate authority"? Just because a cert isn't trusted by default, doesn't mean that it's "self-signed", does it? It could be signed by a centralized CA owned by the users organisation, and I guess that's what they'd need to import here in that case, is it?

`self-signed cert` is relative to whether it's issuer has been included in the trusted Certificate Authority bundle for the environment.  Your example is correct and we are generalizing all of these "not trusted by default certificates" under the term `self-signed cert`.  In a perfect environment, this PR only adds support for working with S3 stores in non-production environments

grdryn:
> While I agree TLS is the correct term there is potential confusion with underlying dependencies that still use the term SSL for their public confgs & apis (See https://github.com/opendatahub-io/ai-edge/pull/241#discussion_r1562663813
). In documentation, we could enforce the use of SSL/TLS where both apply or just tls based features/configs that we own completely

I definitely haven't thought about this as much as you have (and don't necessarily fully understand all of the implications). My suggestion was mainly based on having a quick scan at other projects documentation, and how some appear to prefer "TLS certificate" rather than "SSL certificate" or SSL/TLS certificate". I don't have a strong enough opinion, so fine with it being ssl.

test/e2e-tests/README.md
```diff
@@ -36,10 +37,12 @@ The following enviroment variables are required to run the tests. If any of thes
 
 The following enviroment variables are optional. You may still need to set them for the tests to pass depending on your setup. Read [here](../../pipelines/README.md#ai-edge-end-to-end-pipeline) for more context on how to set these up.
 
-- `SELF_SIGNED_CERT` - A path on your local machine to a self signed cert to be used in pipeline
+- `GIT_SELF_SIGNED_CERT` - Self-signed cert to be used by git when cloning
+- `S3_SELF_SIGNED_CERT` - Self-signed cert to be used for S3 when pulling models
 - `GO` - Custom Go installation path that is not set in your `PATH`
 
 ```bash
+- `S3_SELF_SIGNED_CERT` - Self signed cert to be used for S3 when pulling models
```
grdryn, line 45:
```suggestion
```

^ this looks like a mistake - it's already listed on line 41, and doesn't seem to make sense here inside this code block

grdryn (approved):
/approve with one little inline comment.

@LaVLaS I think you've tested this, so I'll leave it to you to override the test job and give the lgtm

LaVLaS (approved):
/approve

Everything work as intended (success & expected failure) when a cert is required.  Additionally, existing works flow where a cert is not required and omitted worked as inteded

LaVLaS:
> I definitely haven't thought about this as much as you have (and don't necessarily fully understand all of the implications). My suggestion was mainly based on having a quick scan at other projects documentation, and how some appear to prefer "TLS certificate" rather than "SSL certificate" or SSL/TLS certificate". I don't have a strong enough opinion, so fine with it being ssl.

You are correct on the use of TLS over SSL.  I have no object on using the `TLS` or at minimum `SSL/TLS` in all documentation.  The use in the pipeline was just parity with its use in the [git-clone task](https://artifacthub.io/packages/tekton-task/tekton-catalog-tasks/git-clone#cloning-private-repositories) and [kserve/storage library](https://github.com/opendatahub-io/kserve/blob/v0.11.1.3/python/kserve/kserve/storage/storage.py#L174)

LaVLaS:
/override ci/prow/test-ai-edge
/lgtm

Adding override due to openshift-ci downtime and offline tests were successful for each use case with/without certs and existing functionality was preserved

test/e2e-tests/README.md
jackdelahunt: Created a small follow up https://github.com/opendatahub-io/ai-edge/pull/248

PR #242, opened 2024-04-04
RHOAIENG-436: Add model-name to labels and generateName

Sara4994:
## Description

This PR is a simple one to include model-name as label to the pipelineruns helping in filtering the pipelineruns based on labels and also attach model-name as part of generateName to distinguish the pipelineruns name.

## How Has This Been Tested?

- You can create a new pipelinerun to see the change in name of the pipelinerun
- You can run below command to filter the pipelineruns based on the label
   `oc get pipelineruns -l model-name=bike-rentals-auto-ml`

Screenshot 2024-04-04 at 11 48 42 PM

## Merge criteria:

### DIFF
--- a/acm/odh-edge/apps/tensorflow-housing-app/kustomization.yaml
+++ b/acm/odh-edge/apps/tensorflow-housing-app/kustomization.yaml
@@ -19,6 +19,7 @@ namePrefix: tensorflow-housing-app-
 
 commonLabels:
   app: tensorflow-housing-app-1
+  model-name: tensorflow-housing
 
 configMapGenerator:
 - name: test-data
@@ -31,7 +32,6 @@ patches:
     kind: Deployment
     metadata:
       name: model-1
-      model-name: tensorflow-housing
     spec:
       template:
         spec:
--- a/acm/odh-edge/apps/tensorflow-housing-app/kustomization.yaml
+++ b/acm/odh-edge/apps/tensorflow-housing-app/kustomization.yaml
@@ -31,6 +31,7 @@ patches:
     kind: Deployment
     metadata:
       name: model-1
+      model-name: tensorflow-housing
     spec:
       template:
         spec:
--- a/pipelines/tekton/aiedge-e2e/aiedge-e2e.bike-rentals.pipelinerun.yaml
+++ b/pipelines/tekton/aiedge-e2e/aiedge-e2e.bike-rentals.pipelinerun.yaml
@@ -4,7 +4,6 @@ metadata:
   labels:
     tekton.dev/pipeline: aiedge-e2e
     model-name: bike-rentals-auto-ml
-
   generateName: aiedge-e2e-bike-rentals-auto-ml-
 spec:
   params:
--- a/pipelines/tekton/aiedge-e2e/aiedge-e2e.bike-rentals.pipelinerun.yaml
+++ b/pipelines/tekton/aiedge-e2e/aiedge-e2e.bike-rentals.pipelinerun.yaml
@@ -3,7 +3,9 @@ kind: PipelineRun
 metadata:
   labels:
     tekton.dev/pipeline: aiedge-e2e
-  generateName: aiedge-e2e-
+    model-name: bike-rentals-auto-ml
+
+  generateName: aiedge-e2e-bike-rentals-auto-ml-
 spec:
   params:
   - name: model-name
--- a/pipelines/tekton/aiedge-e2e/test-data/bike-rentals-test-data-cm.yaml
+++ b/pipelines/tekton/aiedge-e2e/test-data/bike-rentals-test-data-cm.yaml
@@ -3,7 +3,7 @@ apiVersion: v1
 metadata:
   name: bike-rentals-test-data
   labels:
-    model-name: bike-rentals-test-data
+    model-name: bike-rentals-auto-ml
 
 data:
   data.json: |+
--- a/pipelines/tekton/aiedge-e2e/test-data/bike-rentals-test-data-cm.yaml
+++ b/pipelines/tekton/aiedge-e2e/test-data/bike-rentals-test-data-cm.yaml
@@ -2,6 +2,7 @@ kind: ConfigMap
 apiVersion: v1
 metadata:
   name: bike-rentals-test-data
+  model-name: bike-rentals-test-data
 
 data:
   data.json: |+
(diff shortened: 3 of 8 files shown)


pipelines/tekton/aiedge-e2e/aiedge-e2e.bike-rentals.pipelinerun.yaml
```diff
@@ -3,7 +3,9 @@ kind: PipelineRun
 metadata:
   labels:
     tekton.dev/pipeline: aiedge-e2e
-  generateName: aiedge-e2e-
+    model-name: bike-rentals-auto-ml
+
```
MarianMacik, line 7:
```suggestion
```

MarianMacik (commented):
The jira also mentions model version, should this be also reflected in the labels section?

Sara4994:
As per recent [discussion](https://issues.redhat.com/browse/RHOAIENG-436?focusedId=24486865&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-24486865) on the JIRA, added labels to inference container test Deployment as well.

acm/odh-edge/apps/tensorflow-housing-app/kustomization.yaml
```diff
@@ -31,6 +31,7 @@ patches:
     kind: Deployment
     metadata:
       name: model-1
+      model-name: tensorflow-housing
```
grdryn, line 34: The deployment already gets an `app: <model name>` label through the kustomization file, here: https://github.com/opendatahub-io/ai-edge/blob/main/acm/odh-edge/apps/bike-rental-app/kustomization.yaml#L10

If we need an additional label for `model-name`, then I suggest we add it there, rather than to the Deployment directly like this.

pipelines/tekton/aiedge-e2e/aiedge-e2e.bike-rentals.pipelinerun.yaml
LaVLaS: The recommendation for model-version was a suggestion IF tekton supported the ability to dynamically add parameters as label `key=value` pairs.  It doesn't look like that exists so we can only provide an example with static values until we have automation that supports creating a pipeline run with the model version as a label value

MarianMacik:
> As per recent [discussion](https://issues.redhat.com/browse/RHOAIENG-436?focusedId=24486865&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-24486865) in the JIRA, added labels to inference container test Deployment as well.

I wasn't a part of the discussion but doesn't the inference container test Deployment mean this?
https://github.com/opendatahub-io/ai-edge/blob/97f36c9c35b6dbd9c8237449a2a17cc8c3931fae/pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml#L223-L273

So it would mean to add labels to this Deployment and not the apps, right?

MarianMacik (approved):
Looks good to me now. I will leave final lgtm word to @grdryn :slightly_smiling_face:

pipelines/tekton/aiedge-e2e/test-data/bike-rentals-test-data-cm.yaml
```diff
@@ -2,6 +2,7 @@ kind: ConfigMap
 apiVersion: v1
 metadata:
   name: bike-rentals-test-data
+  model-name: bike-rentals-test-data
```
grdryn, line 5:
```suggestion
  labels:
    model-name: bike-rentals-test-data
```

pipelines/tekton/aiedge-e2e/test-data/tensorflow-housing-test-data-cm.yaml
```diff
@@ -2,6 +2,7 @@ kind: ConfigMap
 apiVersion: v1
 metadata:
   name: tensorflow-housing-test-data
+  model-name: tensorflow-housing-test-data
```
grdryn, line 5:
```suggestion
  labels:
    model-name: tensorflow-housing-test-data
```

grdryn (changes requested):

grdryn:
@LaVLaS are there any conventions for labels in ODH? Like should the label key have a prefix, like `opendatahub.io/model-name` or something? I think a prefixed label would be a little better.

pipelines/tekton/aiedge-e2e/test-data/bike-rentals-test-data-cm.yaml
grdryn: ^at least I guess this is the intention

LaVLaS:
> @LaVLaS are there any conventions for labels in ODH? Like should the label key have a prefix, like `opendatahub.io/model-name` or something? I think a prefixed label would be a little better.

That is a great suggestion for the global identifier label to separate Pipelines, Tasks, *Runs, Pods, .... that are "owned" by the ai-edge workflow.  If/When we have a component name then the format would be `app.opendatahub.io/<AI EDGE COMPONENTNAME>: true'.  For model specific runtime information that is a being used as a label (model name, version, ....), that falls under a label key that is private/specific to the user so a prefix would not be required, IMHO.

I don't think we should block this PR for that discussion but we should add this label standard to the discussions for creating controller to manage this AI Edge application workflow

pipelines/tekton/aiedge-e2e/test-data/bike-rentals-test-data-cm.yaml
```diff
@@ -2,6 +2,8 @@ kind: ConfigMap
 apiVersion: v1
 metadata:
   name: bike-rentals-test-data
+  labels:
+    model-name: bike-rentals-test-data
```
grdryn, line 6:
```suggestion
    model-name: bike-rentals-auto-ml
```

^this is the model name 🙂

pipelines/tekton/aiedge-e2e/test-data/tensorflow-housing-test-data-cm.yaml
```diff
@@ -2,6 +2,8 @@ kind: ConfigMap
 apiVersion: v1
 metadata:
   name: tensorflow-housing-test-data
+  labels:
+    model-name: tensorflow-housing-test-data
```
grdryn, line 6:
```suggestion
    model-name: tensorflow-housing
```

grdryn (changes requested):

grdryn (approved):
/lgtm

grdryn:
> I don't think we should block this PR for that discussion but we should add this label standard to the discussions for creating controller to manage this AI Edge application workflow

Sure 👍

LaVLaS:
/approve
/override ci/prow/test-ai-edge

Ignoring test failures due to extended openshift-ci environment provision issues that are preventing successful test runs

PR #246, opened 2024-04-09
RHOAIENG-4146 - Add support to import a model from the ODH Model Registry via Model Build Pipeline

biswassri:
Adding support to use model-registry and fetch model properties from it when enabled and to continue supporting the existing pipeline without MR deployed.
JIRA: [RHOAIENG-4146](https://issues.redhat.com//browse/RHOAIENG-4146)

## Description

Added a Task that fetches properties stored in model registry when the parameter `model-registry-enabled` is true. So far it only fetches the model source details (git repo, revision etc.) and not the containerfile details. 

The added task `fetch-data` runs always, when `model-registry-enabled` is set to `true` then the Task connects to MR and fetches all the model related properties from the MR specified in the `model-registry-hostname`.  The fetch-git and fetch-s3 tasks runs based on the results of fetch-model from this task.  When set to `false` it passes on the properties from the PipelineRun file. 

Below is a sample curl command used to save the model properties in the MR. 

```
curl -X 'POST' \
  "$MR_HOSTNAME/api/model_registry/v1alpha3/model_versions" \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "description": "Sample for git workflow",
  "name": "2",
  "registeredModelID": "15",
  "customProperties": {
    "fetch-model": {
      "string_value": "git",
      "metadataType": "MetadataStringValue"
    },
    "git-model-repo": {
      "string_value": "https://github.com/opendatahub-io/ai-edge.git",
      "metadataType": "MetadataStringValue"
    },
    "modelRelativePath": {
      "string_value": "pipelines/models",
      "metadataType": "MetadataStringValue"
    },
    "model-dir": {
      "string_value": "tf2model",
      "metadataType": "MetadataStringValue"
    },
    "git-model-revision": {
      "string_value": "main",
      "metadataType": "MetadataStringValue"
    }
  }
}'
```

## How Has This Been Tested?

- Deploy Model Registry in cluster using the `serviceRoute` set to enabled state. 
- Once this option is set to enabled we add the MR Route under `model-registry-hostname` parameter to the pipeline run file
- After MR operator is deployed and populated with model parameters that are require we run the pipelineRun commands: 
`$oc apply -k tekton/aiedge-e2e/
 $oc create -f tekton/aiedge-e2e/aiedge-e2e.tensorflow-housing.pipelinerun.yaml`
 -  The remaining workflow remains the same only the pipeline run becomes shorter

 

You'd expect the below green pipelineRun :
Screenshot 2024-04-17 at 10 55 58 PM

## Merge criteria:


jackdelahunt:
So was just looking at this and wanted to get a better understanding of what is happening.

Fetching from the model registry is now the default and the results from the model registry (info about model in s3 or git) are passed to the task that will then get the model? Is that correct?

Is it expected that we only use model registry now?

pipelines/tekton/aiedge-e2e/tasks/model-registry-download.yaml
```diff
@@ -0,0 +1,110 @@
...
+    type: string
+  - name: git-model-revision
+    type: string
+  steps:
+  - name: download-model-registry
+    image: registry.access.redhat.com/ubi9/ubi
+    script: |
+      #!/usr/bin/env bash
+
+      if [ "$(params.model-registry-enabled)" == "true" ]; then
```
jackdelahunt, line 54: So what is the difference between model-registry-enabled being true or false? Are the fetch tasks (git, s3) supposed to work either way?

biswassri: @jackdelahunt  Thanks for the comment. I updated the Readme to reflect the answer to this. Hope that helps.

pipelines/tekton/aiedge-e2e/tasks/model-registry-download.yaml
```diff
@@ -0,0 +1,115 @@
...
+  - name: model-dir
+    type: string
+  - name: git-model-revision
+    type: string
+  steps:
+  - name: download-model-registry
+    image: registry.access.redhat.com/ubi9/ubi
+    script: |
+      #!/usr/bin/env bash
+
```
grdryn, line 53:
```suggestion

set -Eeuo pipefail

```

pipelines/tekton/aiedge-e2e/tasks/model-registry-download.yaml
```diff
@@ -0,0 +1,115 @@
...
+          jq -r --arg MODEL_PARAM "modelRelativePath" '.customProperties[$MODEL_PARAM].string_value')
+          echo -n "$MODEL_RELATIVE_PATH" | tee $(results.modelRelativePath.path)
+
+          export S3-BUCKET-NAME=$(curl --silent -X 'GET' \
+          "$(params.model-registry-hostname)/api/model_registry/v1alpha3/model_versions/$MODEL_VERSION_ID" -H 'accept: application/json' | \
+          jq -r --arg MODEL_PARAM "s3-bucket-name" '.customProperties[$MODEL_PARAM].string_value')
+          echo -n "$S3-BUCKET-NAME" | tee $(results.s3-bucket-name.path)
+        fi
+        else
+          echo -n "Fetching params from PipelineRun"
```
grdryn, line 102:
```suggestion
          echo -n "Using params from PipelineRun"
```

pipelines/tekton/aiedge-e2e/tasks/model-registry-download.yaml
```diff
@@ -0,0 +1,115 @@
...
+          echo -n "$(params.model-dir)" | tee $(results.model-dir.path)
+          echo -n "$(params.modelRelativePath)" | tee $(results.modelRelativePath.path)
+          echo -n "$(params.s3-bucket-name)" | tee $(results.s3-bucket-name.path)
+      fi  
+      
+  workspaces:
+  - description: The workspace for the downloaded model uris.
+    name: workspace
+  - description: The S3 secret.
+    name: s3-secret
```
grdryn, line 115: Are either of these workspaces actually used here?

pipelines/tekton/aiedge-e2e/aiedge-e2e.tensorflow-housing.pipelinerun.yaml
```diff
@@ -9,11 +9,15 @@ spec:
   - name: model-name
     value: tensorflow-housing
   - name: model-version
-    value: "1"
+    value: "2"
   - name: s3-bucket-name
     value: rhoai-edge-models
+  - name: model-registry-hostname
+    value: "modelregistry-sample-http-default.apps.srbiswas-cluster2.dev.datahub.redhat.com"
```
jackdelahunt, line 16: If you are doing another pipeline run make sure to just change the host name here

pipelines/README.md
```diff
@@ -117,6 +117,12 @@ Create a copy of the file(s) below to include the required credentials for acces
     # Linking secret is needed only if the secret from the Robot account hasn't been already applied and linked in the previous step
     $ oc secret link pipeline credentials-image-registry
     ```
+### Setup ODH Model Registry for Storing Models
```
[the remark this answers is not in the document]
biswassri: The one in this PR is the link to the MR operator repo which is changing fast. 
@devguyio Maybe we would need to update the CLI with the link for the MR installations because anyone using the latest changes of MR operators would see errors?

pipelines/tekton/aiedge-e2e/aiedge-e2e.tensorflow-housing.pipelinerun.yaml
```diff
@@ -9,11 +9,15 @@ spec:
   - name: model-name
```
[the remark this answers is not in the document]
biswassri: I agree, its a good idea. I'll add a separate pipelinerun example.

pipelines/tekton/aiedge-e2e/aiedge-e2e.tensorflow-housing.pipelinerun.yaml
biswassri: Yeah I was planning to remove it, but since we don't have any shared cluster, should i just leave it as `<insert model-registry-hostname` in the example pipelinerun?

pipelines/tekton/aiedge-e2e/tasks/model-registry-download.yaml
```diff
@@ -0,0 +1,115 @@
...
+  - name: download-model-registry
+    image: registry.access.redhat.com/ubi9/ubi
+    script: |
+      #!/usr/bin/env bash
+
+      if [ "$(params.model-registry-enabled)" == "true" ]; then
+        yum install jq -y
+
+        export MODEL_ID=$(curl --silent -X 'GET' \
+        "$(params.model-registry-hostname)/api/model_registry/v1alpha3/registered_models?" -H 'accept: application/json' | \
```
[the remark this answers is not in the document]
grdryn: > My thinking was that since the MR would be deployed on the same 'local' cluster as that of the pipeline, so it should/would have access to the endpoint.

That will _probably_ typically be the case, although I do wonder if there will be cases where the pipelines are run on a different cluster than the model registry. I notice that you're using a Route address to access the model registry (rather than a Service address internal to the cluster), which indicates that the model registry will at least be accessible to the outside world.

> I don't think MR have implemented anything with authentication. Do you see anything that I am missing?

Yeah, I'm not sure really. Authentication could be external to the API, using something like the OpenShift OAuth proxy or something else sitting in front of the API.

I'm not sure if it's 100% necessary to be able to handle authentication for this at this point, but it's something that we should at least think about.

@LaVLaS do you have any thoughts about this?

@devguyio have you thought about authentication for the model registry in the CLI at all?

pipelines/tekton/aiedge-e2e/aiedge-e2e.tensorflow-housing.pipelinerun.yaml
jackdelahunt: Yeah just some example value that shows what it should be without actually working.

`modelregistry-sample-http-default.apps.username-cluster.dev.datahub.redhat.com` or something

grdryn:
The PR check failure appears to be as a result of the following error in the PipelineRun:
```
[pod/aiedge-e2e-bike-rentals-auto-ml-k8s75-build-container-image-pod/step-build-and-push] [1/2] STEP 5/6: COPY $MODEL_DIR /opt/app-root/src/model/
[pod/aiedge-e2e-bike-rentals-auto-ml-k8s75-build-container-image-pod/step-build-and-push] Error: building at STEP "COPY $MODEL_DIR /opt/app-root/src/model/": checking on sources under "/workspace/source/model_dir-bike-rentals-auto-ml/bike-rentals-auto-ml": copier: stat: "/model_dir-bike-rentals-auto-ml": no such file or directory
```

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/opendatahub-io_ai-edge/246/pull-ci-opendatahub-io-ai-edge-main-test-ai-edge/1783098369141379072/artifacts/test-ai-edge/test-ai-edge/artifacts/aiedge-e2e-bike-rentals-auto-ml-k8s75/pipelineLogs.txt

pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
```diff
@@ -116,24 +124,54 @@ spec:
...
     params:
     - name: url
-      value: $(params.git-model-repo)
+      value: $(tasks.fetch-model-registry.results.git-model-repo)
     - name: revision
-      value: $(params.git-model-revision)
+      value: $(tasks.fetch-model-registry.results.git-model-revision)
     - name: subdirectory
-      value: /model_dir-$(params.model-name)/
+      value: model_dir-$(params.model-name)/
```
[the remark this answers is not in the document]
grdryn: Could this be what's causing the PR check to fail?

```
[pod/aiedge-e2e-bike-rentals-auto-ml-k8s75-build-container-image-pod/step-build-and-push] [1/2] STEP 5/6: COPY $MODEL_DIR /opt/app-root/src/model/
[pod/aiedge-e2e-bike-rentals-auto-ml-k8s75-build-container-image-pod/step-build-and-push] Error: building at STEP "COPY $MODEL_DIR /opt/app-root/src/model/": checking on sources under "/workspace/source/model_dir-bike-rentals-auto-ml/bike-rentals-auto-ml": copier: stat: "/model_dir-bike-rentals-auto-ml": no such file or directory
```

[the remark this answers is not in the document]
biswassri: It's possible actually, I earlier had seen another error with the path: 
`"Error running git [init //workspace/output/model_dir-tensorflow-housing/]:` hence this change.  But maybe something else was the issue them, I'm trying to reverse this back and test again

pipelines/tekton/aiedge-e2e/tasks/model-registry-download.yaml
biswassri: Sorry I missed this comment earlier. Good catch, actually this is not being used, I'm gonna test with this removed.

pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
```diff
@@ -116,24 +124,54 @@ spec:
...
+    - name: model-registry-hostname
+      value: $(params.model-registry-hostname)
+    - name: model-registry-enabled
+      value: $(params.model-registry-enabled)
+    - name: git-model-repo
+      value: $(params.git-model-repo)
+    - name: git-model-revision
+      value: $(params.git-model-revision)
+    - name: model-dir
+      value: model_dir-$(params.model-name)/
```
MarianMacik, line 142: The source of the issue is this:
```suggestion
    - name: model-dir
      value: $(params.model-dir)
```

This directory is just referencing the directory for the Containerfile to copy the actual model files from. This directory is relative to the `CONTEXT` which for buildah task is: `model_dir-$(params.model-name)/$(tasks.fetch-model-registry.results.modelRelativePath)/$(params.model-name)`
So the 
```
checking on sources under "/workspace/source/model_dir-bike-rentals-auto-ml/bike-rentals-auto-ml": copier: stat: "/model_dir-bike-rentals-auto-ml": no such file or directory
```
actually meant I cannot find `/model_dir-bike-rentals-auto-ml` under `/workspace/source/model_dir-bike-rentals-auto-ml/bike-rentals-auto-ml`, which now makes complete sense but initially is confusing.

pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
```diff
@@ -116,24 +124,54 @@ spec:
...
+    - name: fetch-model
+      value: $(params.fetch-model)
+    - name: modelRelativePath
+      value: $(params.modelRelativePath)
```
MarianMacik, line 146: Missing:
```suggestion
      value: $(params.modelRelativePath)
    - name: s3-bucket-name
      value: $(params.s3-bucket-name)
```

Also, it would be good to have the parameters in the same order both in the Pipeline here and in the Task.
Another improvement can be to have the `s3-bucket-name` parameter without the default value in the task. This way, the pipeline would most likely fail that the pipeline is not passing the parameter to the task. But once properly linked, it won't be failing because the pipeline parameter `s3-bucket-name` itself has a default value of `""`.

pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
```diff
@@ -116,24 +124,54 @@ spec:
     description: The buildah version used to build the model container image
     value: $(tasks.retrieve-build-image-info.results.buildah-version)
   tasks:
+  - name: fetch-model-registry
```
MarianMacik, line 127: Shouldn't we split into 2 tasks for better clarity? This is both fetching model registry + it also processes data.

Probably it would be better to make this task optional (just run in case the model registry is enabled) and then the subsequent task would either override data from the fetch model registry task or use the pipeline as the source. I get that this would mean to have double the amount of params as one set will be for pipelineRun and the second set would be for parameter values coming from the fetch model registry task. But I would still find it more readable.

Another option would be to name this task better, something like "fetch-model-location-params" or something like that. Because this is what it does, fetching the values either from the registry or from the pipeline run.

MarianMacik (commented):
Initial review.

Also, is this somehow aligned with the CLI work? Currently we have 2 approaches to Model registry:
* CLI - where we store all pipeline params in the model registry
* This - where we get model location from the model registry during the run of the pipeline

@LaVLaS maybe?

pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
biswassri: > Shouldn't we split into 2 tasks for better clarity? This is both fetching model registry + it also processes data.
So IMO it's just fetch the data from the model registry so I don't know how to break this into 2 tasks. There is no connection steps required there is just curl commands for the data. 

I like the override data part for the pipeline, but I don't think that would be syntactically allowed by Tekton. I can't think of way to configure tekton pipeline as to pass params.xyz when model-registry enabled OR task.results.xyz when model-registry enabled.  @grdryn  WDYT?  

We could rename the task but I wanted name to specify that it's involving MR. @MarianMacik  Does that make sense?

MarianMacik: @biswassri For that override, we would need to have 2 set of input params for the task which would decide which data to use (either from the model registry or from the pipeline). So if the first task was run (which fetched the model registry), the set of parameters coming from that task's output would be filled in. Then the "decision" task would check if these params are set and use that one as its output or it would just take the values from the pipeline run input params and set it as the output of the task. But as I said, this would mean to have input parameters of that "decision" task doubled as one set would be coming from the model registry task (in case it was not run, it would be empty) and the other set would be coming from the pipeline run params. And the "decision" task would decide which values to put as the output (which will be then used as input for the s3/git task), either model registry values or the pipeline run values.


But this puts unnecessary overhead, I think. So just a renaming might be better. `fetch-model-registry-or-pipelinerun-params` or something similar, maybe?

pipelines/tekton/aiedge-e2e/tasks/model-registry-download.yaml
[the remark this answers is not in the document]
biswassri: @grdryn, I reached out to the MR channel about this question here: https://redhat-internal.slack.com/archives/C05QB0A582E/p1714510019703239. 
I feel the istio part of it is optional so we might be okay temporarily or I might be wrong :)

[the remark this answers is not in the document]
grdryn: I think it's fine for this PR to leave it as-is, like you say. I think we should have a follow-up story to support authentication on it though. WDYT? If you agree, could you create that in Jira?

pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
jackdelahunt: 👍  agree for sure

I think this has been a persistent issue in the repo for a while. Params being out of order or the case we use to be different. I think @biswassri you can fix that here but also we can have a clean up PR for this.

pipelines/tekton/aiedge-e2e/tasks/model-registry-download.yaml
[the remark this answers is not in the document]
biswassri: I agree, other this PR gets too complex and a lot of changes in one. Created the Jira 6703.

pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
biswassri: @MarianMacik My intention with that with this task and that hopefully we would eventually move out of the params from pipelinerun and only use MR for params. So in that case this would be the startpoint of the pipeline always. 

I was thinking of naming this fetch-model-params eventually we can implement ^what you are suggesting and this would be that decision task. WDYT?

MarianMacik: @biswassri Agree. `fetch-model-params` sounds better.

pipelines/tekton/aiedge-e2e/tasks/model-registry-download.yaml
biswassri: @grdryn It turns out we need these workspaces and the pipeline fails without them with the below error: `workspace binding "s3-secret" does not match any declared workspace`

pipelines/tekton/aiedge-e2e/tasks/model-registry-download.yaml
```diff
@@ -0,0 +1,116 @@
...
+          "$(params.model-registry-hostname)/api/model_registry/v1alpha3/model_versions/$MODEL_VERSION_ID" -H 'accept: application/json' | \
+          jq -r --arg MODEL_PARAM "modelRelativePath" '.customProperties[$MODEL_PARAM].string_value')
+          echo -n "$MODEL_RELATIVE_PATH" | tee $(results.modelRelativePath.path)
+
+          export S3-BUCKET-NAME=$(curl --silent -X 'GET' \
+          "$(params.model-registry-hostname)/api/model_registry/v1alpha3/model_versions/$MODEL_VERSION_ID" -H 'accept: application/json' | \
+          jq -r --arg MODEL_PARAM "s3-bucket-name" '.customProperties[$MODEL_PARAM].string_value')
+          echo -n "$S3-BUCKET-NAME" | tee $(results.s3-bucket-name.path)
+        fi
+        else
```
grdryn, line 103:
```suggestion
      else
```

^I think this should be at the same level of indentation as the outer if condition, rather than the inner one just above it. In other words, it's odd to see the following two words at the same level of indentation:

```bash
fi
else
```

pipelines/tekton/aiedge-e2e/tasks/model-registry-download.yaml
grdryn: I'm not suggesting that they be removed from the Pipeline altogether, just from these tasks, if they're not being used within this task. See the `check-model-and-containerfile-exists.yaml` task for example: the `s3-secret` workspace is neither used nor declared.

pipelines/README.md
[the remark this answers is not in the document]
LaVLaS: Unless there is some custom configuration for model registry, I agree that we should defer to the instruction in the model-registry-operator repo.  If we notice any issues with failed deployments then it would be best for the project to file bugs / PRs to improve their instructions then us owning a duplicate set of instructions.

pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
```diff
@@ -116,24 +124,54 @@ spec:
...
+      value: $(params.model-registry-enabled)
+    - name: git-model-repo
+      value: $(params.git-model-repo)
+    - name: git-model-revision
+      value: $(params.git-model-revision)
+    - name: model-dir
+      value: $(params.model-dir)
+    - name: fetch-model
+      value: $(params.fetch-model)
+    - name: modelRelativePath
```
MarianMacik, line 145: Just repeating, so it doesn't get lost. This is still not fixed:

> Missing:
> ```suggestion
>       value: $(params.modelRelativePath)
>     - name: s3-bucket-name
>       value: $(params.s3-bucket-name)
> ```
> 
> 
> 
> Also, it would be good to have the parameters in the same order both in the Pipeline here and in the Task.
> Another improvement can be to have the `s3-bucket-name` parameter without the default value in the task. This way, the pipeline would most likely fail that the pipeline is not passing the parameter to the task. But once properly linked, it won't be failing because the pipeline parameter `s3-bucket-name` itself has a default value of `""`.



Therefore, I would:
1. Make the order of params here and in the task definition same.
2. Remove any default values for the task params, so if any of the param is not set by the pipeline, it will fail.
3. Add the `s3-bucket-name` parameter here as it is still missing.

pipelines/tekton/aiedge-e2e/tasks/model-registry-download.yaml
```diff
@@ -0,0 +1,116 @@
...
+metadata:
+  name: model-registry-download
+spec:
+  description: This Task can be used to fetch a model properties from model registry when enabled or properties passed by pipelinerun 
+  params:
+  - name: model-name
+    type: string
+  - name: model-version
+    type: string
+  #Values to be passed by pipeline run when model-register is disabled
```
MarianMacik, line 12:
```suggestion
  #Values to be passed by pipeline run when model registry is disabled
```

pipelines/tekton/aiedge-e2e/aiedge-e2.model-registry.tensorflow-housing.pipelinerun.yaml
```diff
@@ -0,0 +1,56 @@
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  labels:
+    tekton.dev/pipeline: aiedge-e2e
```
MarianMacik, line 5:
```suggestion
    tekton.dev/pipeline: aiedge-e2e
    model-name: tensorflow-housing
```

pipelines/tekton/aiedge-e2e/aiedge-e2.model-registry.tensorflow-housing.pipelinerun.yaml
```diff
@@ -0,0 +1,56 @@
...
+  - name: model-version
+    value: "2"
+  - name: s3-bucket-name
+    value: rhoai-edge-models
+  - name: model-registry-hostname
+    value: "modelregistry-sample-http-default.apps.srbiswas-cluster2.dev.datahub.redhat.com"
+  - name: git-containerfile-repo
+    value: https://github.com/opendatahub-io/ai-edge.git
+  - name: model-registry-enabled
+    value: "true"
```
MarianMacik, line 20: Just an idea for better readability, try to move this model-registry param to the model-registry-hostname, so it is grouped.

pipelines/tekton/aiedge-e2e/aiedge-e2.model-registry.tensorflow-housing.pipelinerun.yaml
```diff
@@ -0,0 +1,56 @@
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  labels:
+    tekton.dev/pipeline: aiedge-e2e
+  generateName: aiedge-e2e-
+spec:
+  params:
```
MarianMacik, line 8: The list of provided param values is not complete, therefore, the pipeline run is not runnable. Parameters without default values must be provided and I have identified that these are missing:
* modelRelativePath
* fetch-model

pipelines/tekton/aiedge-e2e/tasks/model-registry-download.yaml
```diff
@@ -0,0 +1,116 @@
...
+        jq -r --arg MODEL_NAME "$(params.model-name)" '.items[] | select(.name == $MODEL_NAME).id')
+        
+        export MODEL_VERSION_ID=$(curl -X 'GET' \
+        "$(params.model-registry-hostname)/api/model_registry/v1alpha3/registered_models/$MODEL_ID/versions?" -H 'accept: application/json' | \
+        jq -r --arg MODEL_VERSION "$(params.model-version)" '.items[] | select(.name == $MODEL_VERSION).id')
+        
+        echo -n "MODEL ID is $MODEL_ID"
+        echo -n "MODEL VERSION ID is $MODEL_VERSION_ID"
+
+        if [[ ! -z "$MODEL_ID" ]]; then
```
LaVLaS, line 70: Assuming that an empty `MODEL_ID` is an unrecoverable failure, we should exit with a non-zero error code to prevent any further execution of the pipeline steps since the model information is missing

pipelines/tekton/aiedge-e2e/tasks/model-registry-download.yaml
```diff
@@ -0,0 +1,116 @@
...
+          echo -n "fetching parameters from model-registry"
+
+          export FETCH_MODEL=$(curl --silent -X 'GET' \
```
LaVLaS, line 73: Since we are running a GET against the same endpoint to save values from the response, can we just cache the json response after the first call and use that for all of the variable assignments?  

If we expect an update to the values between each query then there should be a comment to show why so that we don't optimize the calls in a future update.

pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
```diff
@@ -43,6 +43,14 @@ spec:
     type: string
     description: The Git ref to use when cloning the git repo with the saved model
     default: "main"
+  - name: model-registry-hostname
+    type: string
+    default: ""
+    description: The servicename route for model registry
```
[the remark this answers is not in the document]
biswassri: @grdryn My idea of calling it a `servicename` route was that the user would know where or which endpoint to pick-up from the Openshift UI console. In my understanding model registry probably has one route to the db, so I would agree that either is fine.  WDYT?

pipelines/tekton/aiedge-e2e/aiedge-e2.model-registry.tensorflow-housing.pipelinerun.yaml
```diff
@@ -0,0 +1,56 @@
...
+  - name: model-name
+    value: tensorflow-housing
+  - name: model-version
+    value: "2"
+  - name: s3-bucket-name
+    value: rhoai-edge-models
+  - name: model-registry-hostname
+    value: "modelregistry-sample-http-default.apps.srbiswas-cluster2.dev.datahub.redhat.com"
```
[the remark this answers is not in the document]
biswassri: Yes, this is an oversight. Thanks for catching this :)

pipelines/README.md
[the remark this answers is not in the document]
biswassri: So, leaving this as is. We can open a PR to the CLI readme if needed to reflect the same change.

pipelines/tekton/aiedge-e2e/tasks/model-registry-download.yaml
```diff
@@ -0,0 +1,116 @@
...
+          "$(params.model-registry-hostname)/api/model_registry/v1alpha3/model_versions/$MODEL_VERSION_ID" -H 'accept: application/json' | \
+          jq -r --arg MODEL_PARAM "model-dir" '.customProperties[$MODEL_PARAM].string_value')
+          echo -n "$MODEL_DIR" | tee $(results.model-dir.path)
+
+          export MODEL_RELATIVE_PATH=$(curl --silent -X 'GET' \
+          "$(params.model-registry-hostname)/api/model_registry/v1alpha3/model_versions/$MODEL_VERSION_ID" -H 'accept: application/json' | \
+          jq -r --arg MODEL_PARAM "modelRelativePath" '.customProperties[$MODEL_PARAM].string_value')
+          echo -n "$MODEL_RELATIVE_PATH" | tee $(results.modelRelativePath.path)
+
+          export S3-BUCKET-NAME=$(curl --silent -X 'GET' \
```
LaVLaS, line 98: `-` is not allowed in bash shell variable names.  You'll need to change this to `_`

```suggestion
          export S3_BUCKET_NAME=$(curl --silent -X 'GET' \
```

pipelines/tekton/aiedge-e2e/tasks/model-registry-download.yaml
```diff
@@ -0,0 +1,116 @@
...
+          "$(params.model-registry-hostname)/api/model_registry/v1alpha3/model_versions/$MODEL_VERSION_ID" -H 'accept: application/json' | \
+          jq -r --arg MODEL_PARAM "s3-bucket-name" '.customProperties[$MODEL_PARAM].string_value')
+          echo -n "$S3-BUCKET-NAME" | tee $(results.s3-bucket-name.path)
```
LaVLaS, line 101: See above for variable name change

```suggestion
          echo -n "$S3_BUCKET_NAME" | tee $(results.s3-bucket-name.path)
```

LaVLaS:
Adding a reply for historical reference.
> Also, is this somehow aligned with the CLI work? Currently we have 2 approaches to Model registry:
> 
> * CLI - where we store all pipeline params in the model registry

There was a discussion about adding a CLI to support this entire workflow but the scope of the work requires collaboration with the model registry in order to differentiate from the functionality included in the [tekton cli](https://tekton.dev/docs/cli/).  

> * This - where we get model location from the model registry during the run of the pipeline

From an offline conversation with @biswassri about https://github.com/opendatahub-io/ai-edge/pull/246#discussion_r1617421656, the intent was to use the model-registry as an abstraction layer for retrieving the model storage information.  Long term, I think this approach is the better solution as we want to have tasks that can be used to fetch from git or s3 directly with the option of putting some type of registry abstraction task in front that can retrieve the storage information

LaVLaS:
/hold

I'm putting this PR on hold while @biswassri adds changes based on recent comments

pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
biswassri: Done all three steps. Thanks for catching this @MarianMacik

biswassri:
@LaVLaS, @MarianMacik @grdryn I have addressed all the comments (hopefully) and made the changes. Please let me know if I've missed any and if we are good to merge.

biswassri:
As discussed offline with @LaVLaS, this PR will interfere and complicate the refractor that is planned as a part of RHOAISTRAT 234. Hence closing out this PR. However, I will not delete this branch from my private repo and will have the changes incase we decide to revisit this in the future.

PR #250, opened 2024-04-18
Don't use PR branch for rehearsals in the openshift/release repository

MarianMacik:
## Description
I found out that rehearsals run on the openshift/release repository PR set the value of the  `PULL_NUMBER` env variable of the job to the PR in the openshift/release repository. This naturally leads to a rehearsal job run failing because the pipeline is trying to fetch container files from the PR for ai-edge repository, which does not exist with that `PULL_NUMBER` (the PR exists just in the openshift/release repository). Therefore, for rehearsals we should always fetch the default, e.g. the main branch of the ai-edge repository, so the same approach we already use for nightly jobs.

## How Has This Been Tested?
Locally.

## Merge criteria:

### DIFF
--- a/test/shell-pipeline-tests/common.sh
+++ b/test/shell-pipeline-tests/common.sh
@@ -78,7 +78,7 @@ function createImageRegistrySecret() {
 function usePRBranchInPipelineRunIfPRCheck() {
   local -r PIPELINE_RUN_FILE_PATH=$1
 
-  if [ -n "$PULL_NUMBER" ]; then
+  if [ -n "$PULL_NUMBER" ] && [ "$REPO_NAME" == "ai-edge" ]; then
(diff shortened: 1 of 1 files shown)


MarianMacik:
@grdryn I overrode the job as it is failing still. I am currently working on fixing it in https://github.com/openshift/release/pull/51003 but I found out that there is this issue with rehearsals, so I would like to fix that first. I originally expected that the `PULL_NUMBER` of a completely unrelated repository from the tests point of view (openshift/release repo) wouldn't be set but counter-intuitively it is for rehearsals. Can you give it a `lgtm`?

LaVLaS:
/lgtm

Good catch

grdryn (commented):
/lgtm

(sorry for the delay)

PR #254, opened 2024-04-23
RHOAIENG-2244: Add support for fetching models from S3 based on the modelRelativePath & Adds a universal OpenVino ContainerFile

LaVLaS:
## Description
* Updates the `fetch-model-s3` task to provide basic support for fetching a model from a sub directory under the S3 bucket removing the hardcoded requirement that all model directories be stored at the root of the S3 bucket.
* Adds a universal OpenVino containerfile that can build an inference container image for any supported OpenVino Model format

This provides the ability for users to utilize the [OpenShift AI Fraud Detection tutorial](https://access.redhat.com/documentation/en-us/red_hat_openshift_ai_self-managed/2.8/html/openshift_ai_tutorial_-_fraud_detection_example/index) in the workflow to train and deploy a model using our pipelines

Jira: [RHOAIENG-2244](https://issues.redhat.com/browse/RHOAIENG-2244)

## How Has This Been Tested?
* Followed the RHOAI Fraud Detection example to train a model and used that with minio and our Pipeline
OR
* Fetch the `fraud-detection` model from `s3://rhoai-edge/example-models/fraud-detection`

The `test-model-rest-svc` must be updated to match output based on the `data:` not `prediction:`

The `Continerfile.openvino` to build the inferencing container must be used with this model 

## Merge criteria:
Existing pipeline workflows for `bike-rental` and `tensorflow-housing` are not affected

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -29,6 +29,9 @@ This repo currently contains two examples:
 | **Model (Pipelines)** | `tensorflow-housing`     | `bike-rentals-auto-ml` |
 | **App (ACM)**         | `tensorflow-housing-app` | `bike-rental-app`      |
 
+
+*Additionally*, you can follow the [OpenShift AI tutorial - Fraud detection example](https://access.redhat.com/documentation/en-us/red_hat_openshift_ai_self-managed/2.8/html/openshift_ai_tutorial_-_fraud_detection_example/index) to execute the end to end workflow of training a model from scratch & building the [inferencing container](pipelines/containerfiles/Containerfile.openvino) to deliver to Edge environments.
+
 ### Clusters
 
 > **Note**
--- a/pipelines/containerfiles/Containerfile.openvino
+++ b/pipelines/containerfiles/Containerfile.openvino
@@ -1,0 +1,35 @@
+FROM quay.io/opendatahub/openvino_model_server:stable
+
+# Name of the model that will be used as the based directory of the model files AND passed to openvino executable
+ARG MODEL_NAME
+# Location of the model files within the build context directory
+# Structure is based on OpenVino expectations of <BUILD_CONTEXT>/<MODEL_DIR>/<MODEL_VERSION>
+# See https://docs.openvino.ai/2024/ovms_docs_models_repository.html 
+ARG MODEL_DIR="."
+ARG GRPC_PORT=9090
+ARG REST_PORT=8080
+
+ENV MODEL_NAME=$MODEL_NAME
+ENV GRPC_PORT=$GRPC_PORT
+ENV REST_PORT=$REST_PORT
+
+USER root
+
+# Create the root directory for all models stored in this container
+RUN mkdir /models && chown ovms:ovms /models
+
+# Create /models/MODEL_NAME directory to ensure that --model_name passed to OpenVino matches
+# the directory where we store the model and version
+COPY --chown=ovms:ovms $MODEL_DIR /models/$MODEL_NAME
+
+# Remove the TensorFlow fingerprint.pb file if found as it causes an error in OpenVino when multiple
+#   protobuf(.pb) files  are present in the model directory
+#   See https://www.tensorflow.org/guide/saved_model for info on the TensorFlow model
+# Set permissions so that the container can run locally OR on OpenShift
+RUN find /models -name fingerprint.pb -delete && chmod o+rwX /models/ && chgrp -R 0 /models
+
+EXPOSE $REST_PORT $GRPC_PORT
+
+USER ovms
+
+ENTRYPOINT /ovms/bin/ovms --model_path /models/$MODEL_NAME --model_name $MODEL_NAME --port $GRPC_PORT --rest_port $REST_PORT --shape auto --metrics_enable
--- a/pipelines/containerfiles/Containerfile.openvino
+++ b/pipelines/containerfiles/Containerfile.openvino
@@ -1,10 +1,12 @@
+# Inference container image to support models compatible with OpenVino Model Server
+# OpenVino Support model formats - https://docs.openvino.ai/2024/openvino-workflow/model-preparation.html
 FROM quay.io/opendatahub/openvino_model_server:stable
 
 # Name of the model that will be used as the based directory of the model files AND passed to openvino executable
 ARG MODEL_NAME
 # Location of the model files within the build context directory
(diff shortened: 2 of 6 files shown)


pipelines/containerfiles/Containerfile.openvino
adelton, line 1: What is this for? It does not seem to be used in any of the PipelineRuns and hence not tested by our CI. Does it replace `pipelines/containerfiles/Containerfile.openvino.mlserver.mlflow`.

DEVELOPMENT.md
```diff
@@ -29,6 +29,9 @@ This repo currently contains two examples:
 | **Model (Pipelines)** | `tensorflow-housing`     | `bike-rentals-auto-ml` |
 | **App (ACM)**         | `tensorflow-housing-app` | `bike-rental-app`      |
 
+
+*Additionally*, you can follow the [OpenShift AI tutorial - Fraud detection example](https://access.redhat.com/documentation/en-us/red_hat_openshift_ai_self-managed/2.8/html/openshift_ai_tutorial_-_fraud_detection_example/index) to execute the end to end workflow of training a model from scratch & building the [inferencing container](pipelines/containerfiles/Containerfile.openvino) to deliver to Edge environments.
```
adelton, line 33: If we are going to promote the fraud detection tutorial (which I like), shouldn't we just expand all the relevant places and PipelineRuns to actually lead the reader through the steps, and also have things tested by the CI?

adelton (changes requested):

DEVELOPMENT.md
adelton: This is a maintenance liability.

Is it critical that this goes in in this PR? If not, I'd just leave this change out.

pipelines/containerfiles/Containerfile.openvino
adelton: So why can't this be used for the tensorflow-housing instead of the original one, so that it gets exposed and tested, rather than just being a "random" file in the repo?

LaVLaS: > So why can't this be used for the tensorflow-housing instead of the original one,....

I added the comment in the `Containerfile.openvino.mlserver.mlflow` file to address that issue.  The tensorflow-housing models were developed for an [MLFlow](https://github.com/mlflow/mlflow/tree/ef0f922ca874dd94f9b88e8bfbde2a8cb2c385d2/examples/tensorflow) workflow so I would need to update the directory structure for `tensorflow-housing` in Git|S3 to include the versioning structure required by OpenVino.  Since the original model training code was not written for that workflow, my strong preference is to update the training code to support that structure for reproducibility.

 > so that it gets exposed and tested, rather than just being a "random" file in the repo?

This would be a follow-up PR to bring the Fraud Detection model workflow into our supported workflow.  I included it in this PR because it is required working building our OpenVino inference containers with any number of model versions AND it allows anyone reviewing & testing the PR to use this Containerfile to verify OpenVino functionality

DEVELOPMENT.md
LaVLaS: > Is it critical that this goes in in this PR? If not, I'd just leave this change out.

Which change, this text in the DEVELOPERS doc?  I don't have any particular need to included this the document other than to point people to documented workflow that is supported and fits our E2E scenario training and delivering models to the edge.

 > This is a maintenance liability.

I am curious about this. Is it a maintenance liability because it is based on code outside of this repository?

MarianMacik: I get that us writing here that a user can follow that tutorial and use our pipeline for build might mean that it stops working in the future when we don't test it permanently, however, we can at least say it, that it is not permanently tested by us and it is best effort from our side.

pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
```diff
@@ -255,6 +257,7 @@ spec:
               containers:
               - image: $(params.candidate-image-tag-reference)
                 name: $(params.model-name)-$(params.model-version)
+                imagePullPolicy: Always
```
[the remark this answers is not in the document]
LaVLaS: Correct.  I was stuck in a debuging loop testing updates to the Containerfile that were not propagating to the actual test-deployment pod due to a stale directory being present in the running pod although the build step had renamed the folder.  Ultimately it appeared that the hash was a much older build so the quickest fix was to set it to `Always`.  

Another option would be to use the digest hash from the previous build for these test pods to ensure we are testing the exact build

DEVELOPMENT.md
LaVLaS: These are all good points.  I am going to separate all of the "Fraud Detection Model" specific changes, this DEVELOPERS.md and Containerfile, into a separate story/work since some of the longterm fixes should be aligned with a complete rewrite of the pipeline, tasks and manifests.

> I get that us writing here that a user can follow that tutorial and use our pipeline for build might mean that it stops working in the future when we don't test it permanently,

IMO, this scenario does have a benefit of us dogfooding our own tutorials since this tutorial and official documentation  are maintained by another group working on OpenShift AI.  If we find a bug, it would be good to file it early in the upstream releases.

pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml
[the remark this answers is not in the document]
MarianMacik: Yeah, the digest hash might be a bulletproof option, we can fix it in a separate PR maybe.

pipelines/containerfiles/Containerfile.openvino
adelton: Since documentation to do that is not there, I doubt about the utility of that at this point.

DEVELOPMENT.md
adelton: Just pointing people somewhere with some promise of experiencing the end-to-end example across two repos, without including the relevant steps in our documentation, will lead to people (including myself) to get stuck with "OK, what next?" questions, either because the steps are not trivial or because the other repo has changed significantly so our assumptions in the Containerfile no longer work.

Folks will either abandon the attempt, or they will file issues here (better) or in the https://github.com/rh-aiservices-bu/fraud-detection (worse, maintainers of that repo don't know anything about this work). even issues filed against this repo will be hard to respond to in a timely and effective manner if we don't have the experience with that other repo among the team and in the CI.

LaVLaS:
Closing this as the changes are no longer relevant due to a planned refactor of this repository

PR #259, opened 2024-04-30
Use the oc CLI tool to match the rest of the repo documentation.

adelton:
## Description

Use the oc CLI tool to match the rest of the repo documentation.

## How Has This Been Tested?

Manually.

## Merge criteria:

- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has manually tested the changes and verified that the changes work

PR #233, opened 2024-03-21
Remove orphaned Policy and PlacementBinding references

MarianMacik:
## Description
Follow-up to the #229 which contained some unused references to `Policy` and `PlacementBinding`. Unfortunately kustomize doesn't fail for test patch operation if a resource doesn't exist as can be seen in kubernetes-sigs/kustomize/issues/4379.

## How Has This Been Tested?
Ran kustomize before and after, there was no difference in the output, which means the removed parts were not applied.

## Merge criteria:

- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work

PR #264, opened 2024-05-13
test/acm: Remove Policy from kustomize

MarianMacik:
## Description
While going through https://github.com/opendatahub-io/ai-edge/pull/259/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R31 I found out that the `make` `test-acm-%-generate` target doesn't work because the Policy has been removed as a part of #229. This results in the following error:
`Error: nothing selected by Policy.[noVer].policy.open-cluster-management.io/[noName].[noNs]:metadata.name`
which stems from the replacement's `source` part.

Therefore, this PR removes the Policy targets and sources as well as patch references, which makes the make target functional again.
A similar removal has already been done in https://github.com/opendatahub-io/ai-edge/pull/233.

## How Has This Been Tested?
From the root of the repository run:
```shell
make GIT_REPO_URL="https\://github.com/opendatahub-io/ai-edge" GIT_BRANCH=my-git-branch CUSTOM_PREFIX=custom- CUSTOM_APP_NAMESPACE=custom-tensorflow-housing test-acm-bike-rental-app-generate
```
```shell
make GIT_REPO_URL="https\://github.com/opendatahub-io/ai-edge" GIT_BRANCH=my-git-branch CUSTOM_PREFIX=custom- CUSTOM_APP_NAMESPACE=custom-tensorflow-housing test-acm-tensorflow-housing-app-generate
```
Both should output yaml definitions without throwing an error.

## Merge criteria:

### DIFF
--- a/test/acm/bike-rental-app/kustomization.yaml
+++ b/test/acm/bike-rental-app/kustomization.yaml
@@ -22,15 +22,6 @@ patches:
       value: custom-app-namespace
   target:
     kind: ApplicationSet
-- patch: |-
-    - op: replace
-      path: /spec/policy-templates/0/objectDefinition/spec/object-templates/0/objectDefinition/metadata/name
-      value: custom-app-namespace
-    - op: replace
-      path: /spec/policy-templates/0/objectDefinition/metadata/name
-      value: custom-app-namespace-has-argo-label
-  target:
-    kind: Policy
 
 replacements:
 - source:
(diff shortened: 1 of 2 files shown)


grdryn (approved):
/lgtm

@MarianMacik one thought: given that nobody has noticed this in a couple of months, I wonder if the `test-acm-%-generate` target is actually still useful, or whether we should just remove it?

MarianMacik:
> /lgtm
> 
> @MarianMacik one thought: given that nobody has noticed this in a couple of months, I wonder if the `test-acm-%-generate` target is actually still useful, or whether we should just remove it?

@grdryn That is a good question. I will present the proposal for removal in our scrum or in a channel.

PR #266, opened 2024-05-14
RHOAIENG-6869: Support running tests based on values given to the e2e tests

jackdelahunt:
## Description
This PR aims to make the go e2e test more configurable and more usable for cases in which you want to test but don't have all the requirements for the different types of tests.

Using a `config.json` the values are passed to the e2e tests. Based on the this config the e2e tests may or may not run certain tests, this is set from the `enabled` field in each sub object in the config.
```json
{
  "namespace": "",
  "image_registry_username": "",
  "image_registry_password": "",
  "target_image_tags": [],

  "git_fetch": {
    "enabled": false,
    "container_file_repo": "",
    "container_file_revision": "",
    "container_relative_path": "",
    "model_repo": "",
    "model_relative_path": "",
    "model_revision": "",
    "model_dir": "",
    "self_signed_cert": ""
  },

  "s3_fetch": {
    "enabled": false,
    "aws_secret": "",
    "aws_access": "",
    "region": "",
    "endpoint": "",
    "bucket_name": "",
    "self_signed_cert": ""
  },

  "gitops": {
    "enabled": false,
    "token": "",
    "username": "",
    "repo": "",
    "api_server": "",
    "branch": ""
  }
}

```

For example, if you are only trying to run the main MLOps pipeline it is not needed to give the values for gitops. By setting the `enabled` field you can set the gitops pipeline not to run.

The `config.json` is structured to have all non-specific values at the root (Namespace etc.) and then all optional values for tests in their own sub object, here labelled `s3_fetch` and `git_fetch` as they are the two options for fetching. 

## How Has This Been Tested?
- Update the config.json file in the e2e test folder with the values required, the same values used before should work as normal
- Run `make go-test`
- All tests should pass
- Remove the values each fetch or the gitops fields and see how the e2e tests change their behaviour
- All tests should pass again

## Merge criteria:


pipelines/tekton/aiedge-e2e/templates/credentials-image-registry.secret.yaml.template
```diff
@@ -11,5 +11,6 @@ metadata:
     tekton.dev/docker-0: https://quay.io
 type: kubernetes.io/basic-auth
 stringData:
-  username: {{ IMAGE_REGISTRY_USERNAME }}
-  password: {{ IMAGE_REGISTRY_PASSWORD }}
+  username: "{{ IMAGE_REGISTRY_USERNAME }}"
+  password: "{{ IMAGE_REGISTRY_PASSWORD }}"
```
jackdelahunt, line 15: Added quotes to some of the template files as it was breaking the go json parser. I think this shouldn't break the shell tests but I have not tested it

Makefile
```diff
@@ -33,108 +33,8 @@ endif
...
-endif
-ifndef IMAGE_REGISTRY_PASSWORD
-	$(error IMAGE_REGISTRY_PASSWORD is undefined)
-endif
-ifdef GIT_SELF_SIGNED_CERT
-	oc create configmap git-self-signed-cert --from-file=ca-bundle.crt=${GIT_SELF_SIGNED_CERT}
-endif
-ifdef S3_SELF_SIGNED_CERT
-	oc create configmap s3-self-signed-cert --from-file=ca-bundle.crt=${S3_SELF_SIGNED_CERT}
-endif
```
jackdelahunt, line 71: I still need to re-add this part as part of the go testing setup stage

jackdelahunt:
I requested review but still need to update docs and the change I mentioned above. Waiting to do that as I want to get an opinion on the work so far.

grdryn (commented):
> For example, if you are only trying to run the main MLOps pipeline it is not needed to give the values for gitops. If all of the values in the config for gitops are empty then those tests are not run.
> 
> The config.json is structured to have all non-specific values at the root (NAMESPACE etc.) and then all optional values which will be the conditions for tests in their own sub object, here labelled s3-fetch and git-fetch as they are the two options for fetching.

I'm not very enthusiastic about the idea of having a configuration section filled to have a second implicit meaning of "always run the test(s) that use this section".
For example, just because I've got values specified in the "gitops" section, doesn't mean that I always want to run the gitops tests, right?
Is it still possible to run specific tests (or sets or tests) without taking stuff out of the config file?

Also, regarding the config file, there are two surprising (IMHO :slightly_smiling_face:) choices:
- JSON rather than YAML (or something else)
- CAPS_LOCK for the keys

jackdelahunt:
> Is it still possible to run specific tests (or sets or tests) without taking stuff out of the config file?

Maybe just passing the types of tests you want to run through a env variable.

```
TESTS=gitops,s3-fetch make go-test
```

> JSON rather than YAML (or something else)

Go has built in support for json and yaml so I just picked one

> CAPS_LOCK for the keys

This was a carry over from when they were passed and env variables. But yeah they can change

grdryn:
>> Is it still possible to run specific tests (or sets or tests) without taking stuff out of the config file?
>
> Maybe just passing the types of tests you want to run through a env variable.
>
> `TESTS=gitops,s3-fetch make go-test`

Using tags would be one possibility: https://mickey.dev/posts/go-build-tags-testing/

Another would be to have some sort of `enabled` field in the config, which could be set to false (so that I can just set that, rather than take out the entire gitops block, if I don't want to run the gitops-related tests)

jackdelahunt:
> Using tags would be one possibility: https://mickey.dev/posts/go-build-tags-testing/

I have used these before and they were a hassle to deal with but maybe if I was to look into again it might work well here. Seperaste test files have a build falg behind them. How would you pass this to the compiler from make, in a way that isn't a pain

> Another would be to have some sort of enabled field in the config, which could be set to false

This is for sure a thing I was thinking about, in the config types I even have these alerady just that they are infered from the file instead of being set by it. So I think that could be okay.

Something like this I guess you are saying and have this on each sub object in the config file right?
```
"git-fetch": {
    "enabled": true,
    ...
  },
```

test/e2e-tests/support/config.go
```diff
@@ -0,0 +1,101 @@
...
+	TargetImageTags       []string `json:"TARGET_IMAGE_TAGS"`
+
+	GitFetchConfig GitFetchConfig `json:"git-fetch"`
+	S3FetchConfig  S3FetchConfig  `json:"s3-fetch"`
+	GitOpsConfig   GitOpsConfig   `json:"gitops"`
+
+	Clients         *Clients
+	GitFetchEnabled bool
+	S3FetchEnabled  bool
+	GitOpsEnabled   bool
```
jackdelahunt, line 33: Just about what we were saying, moving these to the `GitFetchConfig ` etc. could work

grdryn:
> How would you pass this to the compiler from make, in a way that isn't a pain

We don't necessarily need to require the tests be run by make, right? I'd keep the default invocation that CI would use to run all tests in the Makefile, but there's a bunch of stuff that `go test` can do that we might often want to do without having make cover every possibility. Pointing people to run `go test` for non-default cases should be fine (and probably IDEs can be configured to do that in different configurations)

jackdelahunt:
> Pointing people to run go test for non-default cases should be fine

Yeah I probably agree I just thought we were trying to do everything through the Makefile.

About using build flags, I feel like relying on compiler flags is something we wouldn't want but now that we are saying we just say calling `go test` directly is expected then maybe it would work

MarianMacik:
> Seperate test files have a build falg behind them. How would you pass this to the compiler from make, in a way that isn't a pain
@jackdelahunt what do you mean by that?

This would be just passing the tags variable content or similar to the go test command.

MarianMacik (commented):
A couple of things to think/discuss:
1. Do we need to have this done using a separate config file? Why not having simple switches for meaningful groups of tests (s3, git, gitops...) as Makefile variables? Or we can utilize naming conventions and then use `go test -run regexp` to include just the needed tests. I am not aware of any functionality/library which would enable us tagging each test function like in Java using annotations. The closes to that are [Kubernetes/Kubebuilder markers](https://book.kubebuilder.io/reference/markers.html ) or [codnect/markers](https://github.com/codnect/markers?tab=readme-ov-file#markers) but these are just for static analysis/code generation apparently.
2. I would prefer to provide test parameters using command line rather than using a file. Maybe we could have default Makefile variables defined in the Makefile and then do the validation using Go as Makefile `ifndef ... endif` can be quite repetitive and combining it with `enabled` flags may make it even worse. It is not that expensive to validate parameters later in Go and not earlier in the Makefile IMHO. This also pairs well with the first idea of having these enabled switches as normal Makefile variables. We would just validate the corresponding params only if the switch is NOT false, otherwise the params tied to that flag won't be validated. Tests would then be checking the enabled flags as they do now.

jackdelahunt:
@MarianMacik 

>  Why not having simple switches for meaningful groups of tests (s3, git, gitops...) as Makefile variables?

I guess this could work.. I was thinking about this as then we would also need to do the validation in make but as from what you said you are saying to pass all the variables in, run setup based on other variables set in make like `TEST_GIT_FETCH=true` 

> I would prefer to provide test parameters using command line rather than using a file

My idea with this was to then be able to save a configuration or setup multiple to be run. For example keep the same config but change the bucket and git repo and use self signed certs. But maybe this just ins't a use case.

Another reason for using a cofig is that it is more clear what variables are used where. I can see a case happening where a person is confused but trying to give values for fields that won't be used. This could probably be helped by good docks and clear naming of course aswell.

MarianMacik:
> @MarianMacik
> 
> > Why not having simple switches for meaningful groups of tests (s3, git, gitops...) as Makefile variables?
> 
> I guess this could work.. I was thinking about this as then we would also need to do the validation in make but as from what you said you are saying to pass all the variables in, run setup based on other variables set in make like `TEST_GIT_FETCH=true`

Validation can be done in Go later. I think it would be more clear to validate everything in Go rather than in Makefile with complicated branching. If we had an expensive setup done before running Go, then it would make sense to have it in Makefile to fail fast but if we delegate this to Go, it won't be seconds/minutes later that we find out some parameter is not filled.
> > I would prefer to provide test parameters using command line rather than using a file
> 
> My idea with this was to then be able to save a configuration or setup multiple to be run. For example keep the same config but change the bucket and git repo and use self signed certs. But maybe this just ins't a use case.
> 
> Another reason for using a cofig is that it is more clear what variables are used where. I can see a case happening where a person is confused but trying to give values for fields that won't be used. This could probably be helped by good docks and clear naming of course aswell.

Makes sense then. However, I would like to retain the possibility to pass variables from command line, i.e. not having to use the config file every time. But that may be just my preference. It's true that for CI it may be possible to have a predefined config file and then we can just reference it as an argument/using a Makefile variable. This way we won't need to run sed every time to overwrite the default config, for example.

jackdelahunt:
> Validation can be done in Go later. I think it would be more clear to validate everything in Go rather than in Makefile

Okay I will work on getting that done then 👍 

>  It's true that for CI it may be possible to have a predefined config file and then we can just reference it as an argument/using a Makefile variable

That's true but if every input is thorugh a env variable then we can just load a `.env` file as our config file if we wanted in the CI.

jackdelahunt:
Also just as another question, when we are selecting which tests to run (git fetch,  gitops etc.) which way is nicer to have those flags.

1. Set a single test type either on or off: `TEST_GIT_FETCH=true`, `TEST_GIT_FETCH=false`
2. A list of tests with names we define and check: `TESTS=gitops,fetch_s3,fetch_git`

CFSNM:
> Also just as another question, when we are selecting which tests to run (git fetch, gitops etc.) which way is nicer to have those flags.
> 
>     1. Set a single test type either on or off: `TEST_GIT_FETCH=true`, `TEST_GIT_FETCH=false`
> 
>     2. A list of tests with names we define and check: `TESTS=gitops,fetch_s3,fetch_git`

I would go with option 1. Having one flag per test-suite is more elegant IMO.

MarianMacik:
> > Validation can be done in Go later. I think it would be more clear to validate everything in Go rather than in Makefile
> 
> Okay I will work on getting that done then 👍
> 
> > It's true that for CI it may be possible to have a predefined config file and then we can just reference it as an argument/using a Makefile variable
> 
> That's true but if every input is thorugh a env variable then we can just load a `.env` file as our config file if we wanted in the CI.

I didn't say it would be through env variables, but through the Makefile variables. It's true that then these would need to be translated to env vars for go test or to cobra flags.



> @MarianMacik just updated the PR to load the environment instead of a config file, here are my thoughts so far after using it a bit.
> 
>     1. With moving setup to go there are now more env variables required. With the current changes that is 26 different environment variables with more coming. Personally this is way to much env variables to get from the command line or your IDE config. It is painful to set this up and I am the one developing it so I am going to have more patience then most

Why are more env variables required when moving setup to go? These variables/params will need to be somewhere anyway because the test suite simply needs this data.


>     2. With all of these options, and with all of them needing to be inputed exactly as expected this gives way too many options to not make a mistakes. When just trying to test the git fetch step I needed to re-run the tests ~5 time just because I was getting things wrong.

I understand, though we don't need to pass every variable every time if we configure some defaults in the test suite.

>     3. This is also made worse by the fact the implementation of this is also a pain and tedious which leads to mistakes also
> 
>     4. I think passing values via the command can be good but in this case I think it is a mess. The alternatives may not be better either possibly but I just want to say I don't think this is a great experience.

In the previous project Kogito operator we had Makefile variables which were then [translated using Makefile itself](https://github.com/apache/incubator-kie-kogito-serverless-operator/blob/main/testbdd/Makefile#L1) to flags with prefix `--`. After that [they were passed to a script](https://github.com/apache/incubator-kie-kogito-serverless-operator/blob/main/testbdd/Makefile#L17), which just "filtered" these parameters to [framework-specific parameters prefixed with `--godog.`](https://github.com/apache/incubator-kie-kogito-serverless-operator/blob/main/hack/run-tests.sh#L255) and to [test parameters prefixed with `--tests.`](https://github.com/apache/incubator-kie-kogito-serverless-operator/blob/main/hack/run-tests.sh#L349). The script was quite big because of many parameters but it worked. [After that, these parameters were passed to the `go test` command](https://github.com/apache/incubator-kie-kogito-serverless-operator/blob/main/hack/run-tests.sh#L389) and were further passed by Cobra. I am not saying we should do it like that, we can just take an inspiration from it, e.g. use Cobra as well.

Also, what did you mean by being it messy when passing from the command line and not messy when using a config file? Did you mean that config file is automatically mapped to a Go struct without having to query environment variable one by one? For command line arguments this can be solved by using Cobra, which will automatically parse and map parameters to respective Go variables, for example.

I am not against the config file, just trying to already think about how it will be used by users. If it is run locally, then a user will override the config file with custom values or use a custom config file. For CI we would either need to override the config file using `sed` or provide a specific CI config file, right?

jackdelahunt:
> Why are more env variables required when moving setup to go? These variables/params will need to be somewhere anyway because the test suite simply needs this data.

Sorry, I didn't actually mean there are more because of go just that I added ones that were missing before that make it much more configureable compared to before

> I understand, though we don't need to pass every variable every time if we configure some defaults in the test suite.

This is true but I think many of these defaults are going to need to be set to whatever the user needs more often then not. So much of the options like names, credentials and build tags are specific to the person running it meaning the default won't cover much use cases. Of course setting them to something the pr-check will use means it will be easier for that atleast and you still have some defaults there

> In the previous project Kogito operator we had Makefile variables which were then [translated using Makefile itself](https://github.com/apache/incubator-kie-kogito-serverless-operator/blob/main/testbdd/Makefile#L1) to flags with prefix --. After that [they were passed to a script](https://github.com/apache/incubator-kie-kogito-serverless-operator/blob/main/testbdd/Makefile#L17), which just "filtered" these parameters to [framework-specific parameters prefixed with --godog.](https://github.com/apache/incubator-kie-kogito-serverless-operator/blob/main/hack/run-tests.sh#L255)

I think I understand whats happening but not sure but I do get the next section.

> use Cobra as well

Would `go test` be needed then, could you run tests based on a command given, this then wraps the choice of tests and options passing into one solution but then you might imagine we are creating the ai-edge cli 2.0 so maybe it is too much.

```
ai-edge-e2e test mlops --fetch-model=s3 --aws-secret=...
```

> Also, what did you mean by being it messy when passing from the command line and not messy when using a config file

I meant two things:
1. Its messy as a developer as you don't get to use the nice encoding stuff from go. 
2. It is messy as a user because for a config file I have an example showing the structure and naming and options in one file that is also used in the process, just looking at the config file (which you will already) gives the info you need compared to a naming convention only structure that needs to be documented somewhere else. My guess is though those makefile flags may have autocomplete as well so maybe not as different.

> I am not against the config file, just trying to already think about how it will be used by users.

There are defiantly pros and cons, I am not sure if a config file is the best idea either. Right now I am just feeling the current solution of passing env vars is not my favouite but maybe some of the others mentioned could work instead of either.

> If it is run locally, then a user will override the config file with custom values or use a custom config file. For CI we would either need to override the config file using sed or provide a specific CI config file, right?

Yes that was my idea, for the CI using sed or a custom config can both work for it so either way is fine. For local my idea was, when you clone the repo you get the config template and then copy it (because of gitconfig and maybe leaking secrets) to create your own with all your needed fields filled in.

MarianMacik:
> 
> > use Cobra as well
> 
> Would `go test` be needed then, could you run tests based on a command given, this then wraps the choice of test to run and its specific options into one solution but then you might imagine we are creating the ai-edge cli 2.0 so maybe it is too much.
> 
> ```
> ai-edge-e2e test mlops --fetch-model=s3 --aws-secret=...
> ```

go test can work with Cobra flags, the same approach was used with the Kogito operator I posted.

> > Also, what did you mean by being it messy when passing from the command line and not messy when using a config file
> 
> I meant two things:
> 
>     1. Its messy as a developer as you don't get to use the nice encoding stuff from go.
> 
>     2. It is messy as a user because for a config file I have an example showing the structure and naming and options in one file that is also used in the process, just looking at the config file (which you will already) gives the info you need compared to a naming convention only structure that needs to be documented somewhere else. My guess is though those makefile flags may have autocomplete as well so maybe not as different.
> 
>     3. It means there is no way to permanently save a particular config and if you do want to retry you need to either go back in your command history or saving them manually.
> 
>     4. Passing raw credentials as flags/env vars  I think is a security issue but maybe its actually okay I am not sure but that I was a thought
> 
> 
> > I am not against the config file, just trying to already think about how it will be used by users.
> 
> There are defiantly pros and cons, I am not sure if a config file is the best idea either. Right now I am just feeling the current solution of passing env vars is not my favouite but maybe some of the others mentioned could work instead of either.
> 
> > If it is run locally, then a user will override the config file with custom values or use a custom config file. For CI we would either need to override the config file using sed or provide a specific CI config file, right?
> 
> Yes that was my idea, for the CI using sed or a custom config can both work for it so either way is fine. For local my idea was, when you clone the repo you get the config template and then copy it (because of gitconfig and maybe leaking secrets) to create your own with all your needed fields filled in.

What did you mean by "gitconfig" here?

OK, I think we can go with config file then. We just need a way how to reference a custom config file, either through environment variables or using a Cobra flag (maybe too complex for one parameter, it would also require to use [TestMain instead of normal test approach](https://medium.com/goingogo/why-use-testmain-for-testing-in-go-dafb52b406bc)).

jackdelahunt:
> What did you mean by "gitconfig" here?

Sorry wrong one meant .gitignore. It was just a comment on the fact that if we want to have an up to date config example we should track that but of course we dont want people using that in case they accidentally commit secrets.

> it would also require to use [TestMain instead of normal test approach](https://medium.com/goingogo/why-use-testmain-for-testing-in-go-dafb52b406bc)

Yep that makes sense, it also looks almost the exact same as the init funtion just with out the testing info passed in. I guess it is better for that but I don't know why they both exist. If I dug deeper maybe the scope at which they are called once for...

https://github.com/opendatahub-io/ai-edge/blob/main/test/e2e-tests/tests/pipelines_test.go#L61-L63

Makefile
```diff
@@ -33,108 +33,8 @@ endif
...
-	S3_BUCKET=${S3_BUCKET} \
-	TARGET_IMAGE_TAGS_JSON=${TARGET_IMAGE_TAGS_JSON} \
-	NAMESPACE=${NAMESPACE} \
-	GIT_SELF_SIGNED_CERT=${GIT_SELF_SIGNED_CERT} \
-	S3_SELF_SIGNED_CERT=${S3_SELF_SIGNED_CERT} \
-	GIT_REPO=${GIT_REPO} \
-	GIT_API_SERVER=${GIT_API_SERVER} \
-	GIT_BRANCH=${GIT_BRANCH} \
-	${GO} test -timeout 30m -shuffle off)
+	(cd test/e2e-tests/tests ${GO} test -timeout 60m -shuffle off)
```
LaVLaS, line 37: This needs `&&` to change directory then run `${GO}`

test/e2e-tests/config.json
```diff
@@ -0,0 +1,34 @@
+{
+  "NAMESPACE": "",
+  "IMAGE_REGISTRY_USERNAME": "",
+  "IMAGE_REGISTRY_PASSWORD": "",
+  "TARGET_IMAGE_TAGS": [],
+
```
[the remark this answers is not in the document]
jackdelahunt: I created a template.config.json which can be copied and have its values filled in. The copy needs to be called config.json for the e2e-tests, and I added the config.json to .gitignore to prevent leaks.

So `cp template.config.json config.json` -> fill in fields -> and git isn't tracking you actual config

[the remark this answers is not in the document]
jackdelahunt: * Also fixed the issue you mentioned

Makefile
```diff
@@ -33,108 +33,8 @@ endif
 GO=go
 GOFLAGS=""
 
-# requires:
-#   AWS_SECRET_ACCESS_KEY       - Secret from AWS
-#   AWS_ACCESS_KEY_ID           - Access key from AWS
-#   S3_REGION                   - Region of the bucket used to store the model
-#   S3_ENDPOINT                 - Endpint of the bucket
```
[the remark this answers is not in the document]
jackdelahunt: Updated the e2e-test README

test/e2e-tests/tests/pipelines_test.go
```diff
@@ -81,74 +80,125 @@ func init() {
...
 	if err != nil {
 		t.Fatal(err.Error())
 	}
 
-	pipelineRunPaths := []string{GitOpsUpdateTensorflowHousingPipelineRunRelativePath, GitOpsUpdateBikeRentalsPipelineRunRelativePath}
+	// the pipeline runs we can run here are based on the fetch models used by the mlops pipeline
+	// because this pipeline depends on the finished pipeline run of the mlops pipeline
+	// -> https://issues.redhat.com/browse/RHOAIENG-4982
+	// there is also a possibilty to split these based on the fetch types like the mlops pipeline
+	var pipelineRunPaths []string
```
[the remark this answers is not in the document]
jackdelahunt: But to get the `buildah-sha` do we not need to run the pipeline before to get it out of the results?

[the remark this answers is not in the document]
MarianMacik: Yes, to get the "fresh" image sha, we need to run it.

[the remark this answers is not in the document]
jackdelahunt: Okay I think I understand now, we need to get the results from the mlops pipeline to get the value for `image-digest`. Even though now the gitops pipeline doesn't depend on it directly for these tests they still do just that now we need to do the magic of getting the sha from the first tests.

The comment
```go
// the pipeline runs we can run here are based on the fetch models used by the mlops pipeline
// because this pipeline depends on the pipelines results as it needs the `image-digest` value
```

Makefile
```diff
@@ -33,108 +33,8 @@ endif
...
-#   IMAGE_REGISTRY_USERNAME     - quay.io username
-#   IMAGE_REGISTRY_PASSWORD     - quay.io password
-#   GIT_SELF_SIGNED_CERT        - Self-signed cert to be used in pipeline by git (optional)
-#   S3_SELF_SIGNED_CERT         - Self-signed cert to be used in pipeline by the script that's used to download the model from S3-compatible object storage (optional)
-#   GIT_TOKEN                   - Auth token used by the git ops pipeline to make a pull request
-#   GIT_USERNAME                - Username linked to the GIT_TOKEN
-go-test-setup:
```
LaVLaS, line 47: You can remove this target from the `.PHONY` line

jackdelahunt: 👍

LaVLaS:
/lgtm

After resolving issues with my config.json, this worked for me with a fresh namespace.

One major issue is that this is not testing against the actual testrun that is created by this script so any failed PipelineRuns that already exist cause the `go test` run to fail.  This isn't a blocker for this implementation since we can follow-up when this repo is refactored.

jackdelahunt:
> One major issue is that this is not testing against the actual testrun that is created by this script so any failed PipelineRuns that already exist cause the `go test` run to fail. This isn't a blocker for this implementation since we can follow-up when this repo is refactored.

Okay I will make a follow up for this, I thought I fixed it but I will give it another shot. I will merge this anyway as the happy path is working fine

jackdelahunt:
The issue @LaVLaS was having is fixed in #271 so you can ignore that issue for this PR atleast

test/e2e-tests/tests/pipelines_test.go
```diff
@@ -81,98 +51,219 @@ func init() {
...
+		if err != nil {
+			t.Fatal(err.Error())
+		}
+
+		var targetPipelineRun *v1.PipelineRun
+		for i := 0; i < len(completedPipelineRuns.Items); i += 1 {
+			if completedPipelineRuns.Items[i].CreationTimestamp.After(TestStartTime) {
+				targetPipelineRun = &completedPipelineRuns.Items[i]
+			}
+		}
```
MarianMacik, line 199: We could also just "remember" the pipelineRunIDs, one for S3 and another one for Git. But I think this is OK too.

test/e2e-tests/tests/pipelines_test.go
```diff
@@ -81,98 +51,219 @@ func init() {
...
+		// created after the tests have started then we can ensure we are getting the correct one
+		completedPipelineRuns, err := config.Clients.PipelineRun.List(ctx, metav1.ListOptions{
+			LabelSelector: p.label,
+		})
+		if err != nil {
+			t.Fatal(err.Error())
+		}
+
+		var targetPipelineRun *v1.PipelineRun
+		for i := 0; i < len(completedPipelineRuns.Items); i += 1 {
```
MarianMacik, line 195:
```suggestion
		for i := 0; i < len(completedPipelineRuns.Items); i++ {
```

test/e2e-tests/tests/pipelines_test.go
```diff
@@ -81,98 +51,219 @@ func init() {
...
+
+		// we need to get the results from the pipeline run created by the ml ops pipeline, we also need
+		// to get the correct one that is for this model. Using the correct label and ensuring it was
+		// created after the tests have started then we can ensure we are getting the correct one
+		completedPipelineRuns, err := config.Clients.PipelineRun.List(ctx, metav1.ListOptions{
+			LabelSelector: p.label,
+		})
+		if err != nil {
+			t.Fatal(err.Error())
+		}
```
MarianMacik, line 192: I think this and the timestamp check should be done first for performance reasons, only if we have a previous MLOps pipeline run should we continue with setting up a GitOps pipeline run. Wdyt?

Also, I think that in case a pipeline is not found, we can just call t.Fail() to not fail the test immediately and then call continue? But ideally this should be in separate tests probably. We can improve it later I think, it is just my observation now. I noticed you tried to fix it in #271 though it won't fail test if one MLOps pipeline fails. I know that the corresponding MLOps test fails before, so the result won't be "PASS" but I think we should still fail the GitOps test because it wasn't able to run properly for both MLOps pipelines. But as I said, this can be handled in that other PR.

MarianMacik (commented):
Just minor fixes and after that I think we can merge.

test/e2e-tests/tests/pipelines_test.go
```diff
@@ -81,98 +51,219 @@ func init() {
...
+	}
+
+	if config.GitFetchConfig.SelfSignedCert != "" {
+		support.MountConfigMapAsWorkspaceToPipelineRun("git-self-signed-cert", "git-ssl-cert", &pipelineRun)
 	}
 
-	err = WaitForAllPipelineRunsToComplete(ctx, &clients)
+	support.SetPipelineRunParam("target-image-tag-references", support.NewArrayParamValue(config.TargetImageTags), &pipelineRun)
+	support.SetPipelineRunParam("git-containerfile-repo", support.NewStringParamValue(config.GitFetchConfig.ContainerFileRepo), &pipelineRun)
+	support.SetPipelineRunParam("git-containerfile-revision", support.NewStringParamValue(config.GitFetchConfig.ContainerFileRevision), &pipelineRun)
```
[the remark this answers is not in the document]
jackdelahunt: Is it all `git-containerfile-*` values, I didn't know this

test/e2e-tests/tests/pipelines_test.go
jackdelahunt: Did this now, I also separated the gitops tests into two now aswell

test/e2e-tests/tests/pipelines_test.go
jackdelahunt: Updated now, they are separated and when looking for that pipeline run if it is failed then the gitops part will fail also now

jackdelahunt:
I closed #271 because those fixes were not needed based on new changes here

test/e2e-tests/README.md
```diff
@@ -6,46 +6,53 @@ A local install of the Go compiler is needed to run the tests. Go version `1.21`
...
+The e2e-tests use a `config.json` to read values passed to it. in the `e2e-tests` directory copy the `template.config.json` to `config.json`. You can now fill in the fields in the `config.json`.
 
-The following enviroment variables are optional. You may still need to set them for the tests to pass depending on your setup. Read [here](../../pipelines/README.md#ai-edge-end-to-end-pipeline) for more context on how to set these up.
+```bash
+cp template.config.json config.json
+```
 
-- `GIT_SELF_SIGNED_CERT` - Self-signed cert to be used by git when cloning
-- `S3_SELF_SIGNED_CERT` - Self-signed cert to be used for S3 when pulling models
+The structure of the `config.json` is in four sections, the top level fields, `git_fetch`, `s3_fetch` and gitops. All fields at the top level are required.
```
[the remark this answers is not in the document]
jackdelahunt: Yes that is true, right now it does not fail

test/e2e-tests/template.config.json
```diff
@@ -0,0 +1,37 @@
+{
+  "namespace": "my-test-namespace",
+  "image_registry_username": "username+myrobot",
+  "image_registry_password": "",
+  "target_image_tags": ["quay.io/username/repo repo:latest"],
```
MarianMacik, line 5: Is this tag correct? It should either be `["quay.io/username/repo", "repo:latest"]` or `["quay.io/username/repo:latest"]`, right?

test/e2e-tests/tests/pipelines_test.go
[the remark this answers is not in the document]
MarianMacik: Yes, even if S3 bucket is used for model, for container files we still use Git. So these parameters need to be set for S3 aproach as well. Together with `containerRelativePath`.

test/e2e-tests/template.config.json
jackdelahunt: Yep it was the first one

jackdelahunt:
> Yes, even if S3 bucket is used for model, for container files we still use Git. So these parameters need to be set for S3 aproach as well. Together with containerRelativePath.

Okay no problem updated now

MarianMacik (approved):
Looks good to me. Nice job!

biswassri:
Overall looks good to me. Adding approve since Marian also approved and to get this merged.

PR #267, opened 2024-05-16
Decouple GitOps Update pipeline from aiedge-e2e

grdryn:
JIRA: https://issues.redhat.com/browse/RHOAIENG-4982

## Description

Before this change, at least one PipelineRun for the specified model name from the aiedge-e2e Pipeline needed to exist on the cluster, to be able to successfully run the GitOps Update Pipeline.

This change instead accepts the image details explicitly as parameters in the GitOps Update Pipeline, making it more robust and applicable to more scenarios.

## How Has This Been Tested?

- [x] Tested that all of the updated pipelinerun examples work as expected.
- [x] Observed that the e2e test passes on this PR.

## Merge criteria:

- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work

PR #268, opened 2024-05-21
Add example trigger showing how to auto-trigger GitOps Update pipeline from MLOps build pipeline

grdryn:
JIRA: https://issues.redhat.com/browse/RHOAIENG-4982

This commit is stacked upon #267, and will show the changes from that PR until it's merged. This is one approach that could be taken–I'll open a separate PR with an alternative approach.

## Description

This example shows how users can choose to optionally trigger the GitOps Update pipeline at the end of a successful run of the MLOps build pipeline.

### Advantages of this approach
- It's optional
- It doesn't require making the main build pipeline extra complex, by needing to know everything about the GitOps update pipeline (it doesn't need to know all of the parameters, or tasks; as other approaches might).
- GitOps Update pipeline is still executable in different ways, and can ideally change without needing to also make changes in the build pipeline

### Disadvantages of this approach
- As it relies on an EventListener, an extra pod needs to be running all the time
- Currently an additional trigger (and possibly additional EventListener, although interceptors might be able to ensure only the appropriate trigger(s) is/are fired on a single EventListener) is needed for each different model image and/or GitOps repo.

## How Has This Been Tested?

- [x] Ran the build pipeline without adding a non-empty value for the new `gitops-update-pipeline-eventlistener-url` param, and verified that the optional task to call the EventListener was skipped
- [x] Ran the build pipeline with a valid webhook URL for the new `gitops-update-pipeline-eventlistener-url` param, and verified that the new optional task was executed, and the GitOps Update pipeline was triggered and completed successfully.

## Merge criteria:


MarianMacik:
> Currently an additional trigger (and possibly additional EventListener, although interceptors might be able to ensure only the appropriate trigger(s) is/are fired on a single EventListener) is needed for each different model image and/or GitOps repo.

I don't think it is needed for each different model image because model image is directly in the `image-registry-repo` parameter, isn't it?
Also, GitOps repository can be parameterized as well, right?
In case a user has some internal organization for different GitOps repositories, they only need to change image registry, digest and GitOps repo URL. Or if everything is in one repository, then the `yq-script` must be changed accordingly.

MarianMacik (commented):
Maybe a small piece of documentation referencing the example trigger could be added. Otherwise I think it looks fine.

grdryn:
> > Currently an additional trigger (and possibly additional EventListener, although interceptors might be able to ensure only the appropriate trigger(s) is/are fired on a single EventListener) is needed for each different model image and/or GitOps repo.
> 
> I don't think it is needed for each different model image because model image is directly in the `image-registry-repo` parameter, isn't it?
> Also, GitOps repository can be parameterized as well, right?
> In case a user has some internal organization for different GitOps repositories, they only need to change image registry, digest and GitOps repo URL. Or if everything is in one repository, then the `yq-script` must be changed accordingly.

If anything other than the `image-registry-repo` and `image-digest` parameters need different values for different models and/or GitOps repos (and I expect that they typically will), then a different trigger should be used. I think that's a cleaner approach than having the (already quite cumbersome) aiedge-e2e pipeline needing to know all of the extra parameters for the GitOps Update pipeline.

MarianMacik:
> If anything other than the `image-registry-repo` and `image-digest` parameters need different values for different models and/or GitOps repos (and I expect that they typically will), then a different trigger should be used. I think that's a cleaner approach than having the (already quite cumbersome) aiedge-e2e pipeline needing to know all of the extra parameters for the GitOps Update pipeline.

I see, so these extra parameters would be hard-coded in the trigger template so the aiedge-e2e pipeline don't need to have these at the start.

PR #269, opened 2024-05-27
An example showing what it would be like to have the aiedge-e2e pipeline duplicate all of the gitops update pipeline tasks

grdryn:
JIRA: https://issues.redhat.com/browse/RHOAIENG-4982

## Description

This shows what it would be like to have the aiedge-e2e pipeline also include all of the tasks from the gitops update pipeline.

There are advantages and disadvantages to this approach:

### Advantages:

- Compared to the EventListener/Trigger approach, no Tekton EventListener pod needs to be running all the time.

### Disadvantages:

- The aiedge-e2e pipeline becomes significantly more complex, and it now needs to know how to do everything that the gitops update pipeline knows how to do.
- If we change anything in the gitops update pipeline, we need to remember to also update that in this pipeline.
- Making it optional (for example, for users who don't us a forge that implements the GitHub PR API) would require a 10th additional parameter

## How Has This Been Tested?

Minimally. If this is the chosen path, I'll spend more time on it

## Merge criteria:


adelton:
Would we be able to simplify the code by defining the most complex tasks as `kind: Task` and then just refer to them from both Pipelines?

As for making things optional, could we use somethings like optional `gitops-git-server` or `gitops-git-token-secret-name` to skip the parts of the Pipeline that the user might not want to run?

adelton:
@MarianMacik Would we be able to do CI for both the e2e and the original Pipeline? Perhaps have the e2e create PR to one git branch (like qe) and then have another pipeline to duplicate that SHA256 in another git branch (like stage, or prod) ...

grdryn:
> Would we be able to simplify the code by defining the most complex tasks as `kind: Task` and then just refer to them from both Pipelines?

That's already how this works: it expects the [gitops-update-pipeline/tasks](https://github.com/opendatahub-io/ai-edge/tree/main/pipelines/tekton/gitops-update-pipeline/tasks) to be rolled out, and both pipelines refer to those (and shared ClusterTasks).

> As for making things optional, could we use somethings like optional `gitops-git-server` or `gitops-git-token-secret-name` to skip the parts of the Pipeline that the user might not want to run?

Yeah the new tasks here could easily be skipped by checking for existence of a value on one or more of the needed parameters.

MarianMacik:
> @MarianMacik Would we be able to do CI for both the e2e and the original Pipeline? Perhaps have the e2e create PR to one git branch (like qe) and then have another pipeline to duplicate that SHA256 in another git branch (like stage, or prod) ...

@adelton Sorry, I have just noticed your comment. Yes, that could be done, of course.

PR #271, opened 2024-06-25
Fix e2e tests failing in edge cases

jackdelahunt:
## Description
This PR improves the e2e tests by fixing some cases in which were not considered before.

1. Pipeline runs from past tests either manual or not were failing the test suite as it was looking for any failed run. This has not been changed as it now looks for specific runs created by the test suite.
2. There was a bug where the gitops pipeline tests were assuming both of the mlops pipeline were enabled. This was breaking if this was not true.

Both of these issues were added in #266 but were not fixed there as that is a long running PR and I want to get that merged on its own.

## How Has This Been Tested?
Testing is done as usual, fill in the values in your `config.json` and all should work.

To test the specific issues try this:
1. Create a test with an intentional problem and let it fail. Fix the issue and re-run the tests and all should be okay
2. Run the Gitops pipeline with only either the git fetch or s3 fetch tests enabled and all should be good aswell

## Merge criteria:


jackdelahunt:
@LaVLaS This is a branch from #266 but just fixes some of the follow needed. The PR should give all needed info

PR #272, opened 2024-07-12
RHOAIENG-9787: Create a manifests directory containing MLops pipeline tasks

jackdelahunt:
## Description
Copies the tasks we use for the MLOps pipeline into a new manifests directory

## Merge criteria:


LaVLaS:
Adding context to the PR from offline conversations:
This is the first step to standardize and organize around a `manifests/` directory with proper organization of individual artifacts that can be deployed according to their functionality.  Supported pipelines and tasks that are core to the Near Edge workflow should be organized logically so that can be deployed as an individual group, just tasks VS piplines+tasks.

@jackdelahunt
Since you are migrating the tasks and updating the shell tests we should also remove them from [pipelines/tekton](pipelines/tekton) and update the existing documentation accordingly.

manifests/kustomization.yaml
```diff
@@ -0,0 +1,10 @@
+kind: Kustomization
+apiVersion: kustomize.config.k8s.io/v1beta1
+commonLabels:
+  app: rhoai-edge-pipelines
+  app.kubernetes.io/part-of: rhoai-edge-pipelines
+resources:
+- tasks/
+- aiedge-e2e.pipeline.yaml
+- gitops-update-pipeline.yaml
```
LaVLaS, line 9: For organzational purposes, I would recommending moving the pipelines to a `manifests/pipelines` folder that can deploy the pipelines and required tasks.

manifests/kustomization.yaml
```diff
@@ -0,0 +1,10 @@
...
+- https://raw.githubusercontent.com/tektoncd/catalog/main/task/github-open-pr/0.2/github-open-pr.yaml
```
LaVLaS, line 10: To control the version of Tekton tasks that are supported for this worklow, we should commit a copy of the versioned TektonHub tasks to this repo with some documentation about the source location and how to update it.

I would recommend creating a tasks folder dedicated to official TektonHub tasks so that it is clear that any task in this folder are sourced from the official TektonHub

jackdelahunt: @LaVLaS are you talking about the task here we are getting from github "tektoncd/catalog/main/task/github-open-pr/0.2/github-open-pr.yaml"?

And if so if I am understanding you, you are saying lets copy this to the repo and just say where it is from and how to update it

LaVLaS: Apologies. My comment was for future work which can be handled in a follow-up task to support versioning of the task/pipeline featureset.  This current PR is focused on migrating the current functionality to manifest directory structure so ignore the comment above

jackdelahunt: No problem, only think blocking this now is review and the docs. Do we want to include the readme changes in this PR or do we want to get this through to open up more work for us to do

LaVLaS:
/approve

Follow-up from offline conversation, @biswassri is going to update the README(s) in a follow-up PR to reflect that change in manifests.

PR #276, opened 2024-07-22
RHOAIENG-2244: Make model location the same for every fetch type

jackdelahunt:
## Description
When cloning from git we clone the whole repo and then refer to the model with `modelRelativePath`. When fetching from s3 we are copying the model directly which means `modelRelativePath` is not needed. This ticket is to make the location of the model after each fetch step the same so the following tasks do not need to change their behaviour based on the tasked used before in the pipeline.

This pr adds a new task to be used after the git fetch step to ensure the location of the model once we have moved on from fetching is in a common location.

There are other work arounds instead of adding a new task but we decided to do this as a first solution.

## How Has This Been Tested?
- Run e2e tests as normal
## Merge criteria:

### DIFF
--- a/manifests/pipelines/aiedge-e2e.pipeline.yaml
+++ b/manifests/pipelines/aiedge-e2e.pipeline.yaml
@@ -161,7 +161,7 @@ spec:
     - name: s3-bucket-name
       value: $(params.s3-bucket-name)
     - name: model-relative-path
-      value: $(params.model-relative-path)
+      value: $(params.modelRelativePath)
     taskRef:
       kind: Task
       name: kserve-download-model


manifests/tasks/check-model-and-containerfile-exists.yaml
```diff
@@ -27,18 +27,10 @@ spec:
 
       set -Eeuo pipefail
 
-      # Check model files exist
-      if [ -n "$(params.modelRelativePath)" ]; then
```
MarianMacik, line 31: In that case the `modelRelativePath` can be removed entirely from this task.

manifests/pipelines/aiedge-e2e.pipeline.yaml
```diff
@@ -209,7 +224,7 @@ spec:
     - name: DOCKERFILE
       value: containerfile_repo/$(params.containerfileRelativePath)
     - name: CONTEXT
-      value: model_dir-$(params.model-name)/$(params.modelRelativePath)/$(params.model-name)
+      value: model_dir-$(params.model-name)/$(params.model-name)
```
MarianMacik, line 229: Also in that case change this decription:
```yaml
- name: model-dir
  type: string
  description: Directory below <modelRelativePath>/<model-name>/ to be used as MODEL_DIR in image build.
```
because we no longer have this directory structure. We eliminated the modelRelativePath.

manifests/tasks/kserve-download-model.yaml
```diff
@@ -21,7 +23,12 @@ spec:
     script: |
       mkdir -p $(workspaces.workspace.path)/model_dir-$(params.model-name)/
 
-      export S3_URL="s3://$(params.s3-bucket-name)/$(params.model-name)"
+      if [ -n "$(params.model-relative-path)" ]; then
+        export S3_URL="s3://$(params.s3-bucket-name)/$(params.model-relative-path)/$(params.model-name)"
+        else
```
MarianMacik, line 28:
```suggestion
      else
```

manifests/pipelines/aiedge-e2e.pipeline.yaml
```diff
@@ -136,7 +136,22 @@ spec:
...
       values: ["git"]
-  # Use kserve-download-model container image to fetch from S3 storage
+  - name: cleanup-git-fetch
+    params:
+    - name: model-name
+      value: $(params.model-name)
+    - name: subdirectory
+      value: model_dir-$(params.model-name)
+    - name: src-model-relative-path
+      value: pipelines/models
```
MarianMacik, line 146: This should be
```suggestion
    - name: src-model-relative-path
      value: $(params.model-relative-path)
```
right?
Not hard-coded.

manifests/pipelines/aiedge-e2e.pipeline.yaml
```diff
@@ -136,7 +136,22 @@ spec:
     - input: "$(params.fetch-model)"
       operator: in
       values: ["git"]
-  # Use kserve-download-model container image to fetch from S3 storage
+  - name: cleanup-git-fetch
```
MarianMacik, line 139: Maybe `move-model-to-root-dir` would be better? Or something like that.

manifests/pipelines/aiedge-e2e.pipeline.yaml
jackdelahunt: Yep my mistake

manifests/pipelines/aiedge-e2e.pipeline.yaml
jackdelahunt: Sure NP, I had no idea what to call it

manifests/tasks/check-model-and-containerfile-exists.yaml
jackdelahunt: 👍

manifests/pipelines/aiedge-e2e.pipeline.yaml
```diff
@@ -136,7 +136,22 @@ spec:
...
       values: ["git"]
-  # Use kserve-download-model container image to fetch from S3 storage
+  - name: move-model-to-root-dir
+    params:
+    - name: model-name
+      value: $(params.model-name)
+    - name: subdirectory
+      value: model_dir-$(params.model-name)
+    - name: src-model-relative-path
+      value: $()
```
MarianMacik, line 146: I think you forgot to fill this with the correct param?
```suggestion
      value: $(params.model-relative-path)
```
BTW these suggestions you can just accept I think and they will be committed directly.

manifests/tasks/kserve-download-model.yaml
```diff
@@ -12,6 +12,8 @@ spec:
     type: string
   - name: s3-bucket-name
     type: string
+  - name: model-relative-path
+    tpye: string
```
MarianMacik, line 16: This is why it failed:
```suggestion
    type: string
```

manifests/tasks/move-model-to-root-dir.yaml
```diff
@@ -0,0 +1,28 @@
...
+  - name: copy-model-to-subdirectory
+    image: registry.access.redhat.com/ubi9/ubi
+    script: |
+      #!/usr/bin/env bash
+
+      set -Eeuo pipefail
+
+      mv $(workspaces.workspace.path)/$(params.subdirectory)/$(params.src-model-relative-path)/$(params.model-name) $(workspaces.workspace.path)/
+      rm -rf $(workspaces.workspace.path)/$(params.subdirectory)
+      mkdir $(workspaces.workspace.path)/$(params.subdirectory)
```
biswassri, line 24: @jackdelahunt you can try `mkdir -p ` and you'll no longer need to rm -rf. It basically throws no error if the directory already exists and then mv overwrites over the existing file.

manifests/pipelines/aiedge-e2e.pipeline.yaml
```diff
@@ -136,7 +136,22 @@ spec:
     - input: "$(params.fetch-model)"
       operator: in
       values: ["git"]
-  # Use kserve-download-model container image to fetch from S3 storage
+  - name: move-model-to-root-dir
+    params:
+    - name: model-name
+      value: $(params.model-name)
+    - name: subdirectory
```
biswassri, line 143: This is confusing as the `git clone` task also has a predefined subdirectory. Maybe we have it as model-directory? Or something else. Just that the subdirectory is where git clone by default clones the repository into.

manifests/tasks/move-model-to-root-dir.yaml
jackdelahunt: No because I am intentionally removing everything from the repo excluding the model, which is moved out before the rm

manifests/pipelines/aiedge-e2e.pipeline.yaml
jackdelahunt: But the subdirectory here and in gitclone mean the same thing

jackdelahunt: For the git-fetch step is what I am saying

biswassri: In that case, you format the `/model_dir-$(params.model-name)/` I know it might be just a formatting change but the consistency would be helpful for others reading this.

manifests/pipelines/aiedge-e2e.pipeline.yaml
```diff
@@ -145,6 +160,8 @@ spec:
       value: $(context.pipelineRun.namespace)
     - name: s3-bucket-name
       value: $(params.s3-bucket-name)
+    - name: model-relative-path
+      value: $(params.model-relative-path)
```
MarianMacik, line 164: My bad, sorry.
```suggestion
      value: $(params.modelRelativePath)
```

MarianMacik (approved):
Looks good to me now. Thanks!
/approve

PR #277, opened 2024-07-24
Delete unused tasks and pipeline definitions and update README

MarianMacik:
## Description

* Removed duplicates and fixed README.
* README will be polished further after the Git Fetch pipeline is merged.
* Unified structure of  directories in `pipelines/tekton`. Now there are `templates` directories for secret templates and `example-pipelineruns` for example pipeline runs.
* Refactored variable names in shell tests to better reflect reality.

## How Has This Been Tested?
PR check.

## Merge criteria:


manifests/kustomization.yaml
```diff
@@ -6,4 +6,3 @@ commonLabels:
 resources:
 - tasks/
 - pipelines/
-- https://raw.githubusercontent.com/tektoncd/catalog/main/task/github-open-pr/0.2/github-open-pr.yaml
```
MarianMacik, line 9: This was moved to `tasks/kustomization.yaml` so it is grouped with other tasks.

jackdelahunt (approved):

biswassri (changes requested):
@MarianMacik Some thoughts: 

1.  I was thinking we could maybe also add some blurb about the RHOAI Operator deployment. Something like: This manifest package will be built and deployed with the RHOAI Operator bundle. (Adding links in the future)
2. The Readme should be moved up into the manifest package because going forward the pipeline definition files would be residing there. So, when one looks for the pipelines the readme should be in the same location. 
3. When/If we move up the Readme, we need to remove the same templates from the Readme. We can have a smaller readme in example-pipelines for user reference.

biswassri:
(For the records) After an offline discussion with @MarianMacik, we will re-visit the Readme contents/location after `git-fetch-pipeline` is merged in.

PR #279, opened 2024-07-29
RHOAIENG-9803-Create a sample Git fetch pipeline in manifests

biswassri:
## Description

A pipeline that does the same as the MLops pipeline but now with just the git fetch task as the fetch method. This change also removes the aiedge-e2e-pipeline from the kustomization file. So the files remains but the pipeline is not gonna get created.

## How Has This Been Tested?

Tested the changes using the below commands
```
oc apply -k manifests/
oc create -f pipelines/tekton/aiedge-e2e/git-fetch-tensorflow-housing.pipelinerun.yaml
```
Screenshot 2024-07-29 at 6 15 11 PM

## Merge criteria:

### DIFF
--- a/manifests/pipelines/git-fetch-pipeline.yaml
+++ b/manifests/pipelines/git-fetch-pipeline.yaml
@@ -1,0 +1,360 @@
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: git-fetch
+spec:
+  params:
+  - name: model-name
+    type: string
+    description: Name of the directory where the model files are stored
+  - default: "1"
+    name: model-version
+    type: string
+    description: Version of the model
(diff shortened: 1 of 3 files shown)


jackdelahunt:
Prow will fail because the `pipelines-test-openvino-tensorflow-housing.sh` needs to be updated to use the new pipeline run. I don't think there should be any other breaking changes for that.

E2e-tests will also need to be updated

biswassri:
Thanks @jackdelahunt. I was just going through them. I referred to your PR for the change points. 👍🏻

biswassri:
Closing out this PR, I did some naming changing incorrectly and the gitlog is getting super complex to resolve. Will replace this with a new PR and same changes.

PR #280, opened 2024-07-30
RHOAIENG-6513: Pipeline won't work if "model-name" parameter contains an underscore

jackdelahunt:
## Description
The name of the deployment we create is based on the model name. If the model name contains characters that deployments do not support (only supports [a-z, 0-9, -, .]) then the deployment fails. These changes convert and unsupported characters into a supported one (-) using sed.

**If you need an example model that contains forbidden characters** please ask I have one that can be provided

## How Has This Been Tested?
- Using e2e tests all should pass with a model that contains a character that does not satisfy [a-z, 0-9, -, .]. I can provide if needed

## Merge criteria:


jackdelahunt:
@biswassri @MarianMacik I want to get opinions on how I did this as I am not sure if it is the best way to do things.

I also think there is an issue with the fact we are defining a lot of the bash scripts changed here in the pipelines meaning they are lost at the task level which is what we are delivering.

biswassri:
@jackdelahunt  I think this should work but just some thoughts: 

1. Can we create the deployment-name once, maybe in the test-container-deploy task or even before and then pass on the same Deployment-name to all the following tasks. If you do declare it in the test-container-deploy task then you can pass it along as a result to the following task?  
2. $DEPLOYMENT_NAME is different from how we are declaring other params. If we are able to achieve point 1. I think this would be taken care of by itself. 

WDYT?

grdryn (commented):
> I also think there is an issue with the fact we are defining a lot of the bash scripts changed here in the pipelines meaning they are lost at the task level which is what we are delivering.

Good insight!
My first thought after going through this was that there are some of these tasks that we don't actually provide some of these tasks (we just use the `openshift-client` cluster task directly in the pipeline, rather than having any custom task of our own), so maybe the onus is on customers to provide valid input scripts when using these tasks in their own pipelines?

We do still have our custom `test-model-rest-svc-task` though, and in that (both before and after this proposed change) we rely on an assumption of what the deployment should be named. Maybe a customer would prefer to make a pipeline that they build from these tasks be concurrently runnable: currently if you run two instances of this pipeline in quick succession with the same inputs, the second to reach this stage might clobber the `Deployment`, etc, from the first, right?
So maybe a better approach might be to accept a deployment name parameter in the `test-model-rest-svc-task`, that we can pass through to it, and it isn't coupled to an assumption. Then here in our example pipelines, maybe we could set the deployment name to be something based on the pipeline ID or something?

These resource names are ephemeral for the lifetime of the pipeline, and are cleaned up before the end, right?

manifests/tasks/test-model-rest-svc-task.yaml
```diff
@@ -25,8 +25,10 @@ spec:
 
         echo "\nCall service:"
 
+        export DEPLOYMENT_NAME=`sed 's/[^a-z0-9.-]/-/g; s/-*$//' <<< $(params.model-name)`-$(params.model-version)
```
MarianMacik, line 28: In this task I would name it as a `SERVICE_NAME` since we are hitting a service endpoint here that just happens to have the same name as the deployment.

MarianMacik (commented):
If we want to have this as a part of the pipeline, which is not going to be shipped to users anyway, I would not duplicate it along the pipeline definition but rather create a task `sanitize-model-name` or similar and use that sanitized model name (which will be the output of the sanitize task) across the pipeline. Then it is up to us whether we just want to use it just for the deployment or for all occurrences of model-name throughout the pipeline.

Probably to have changes as minimal as possible, just use this sanitized name for deployment as this is the only places where it produces errors.

We might as well ignore this as this will be up to users of the pipelines to sanitize their inputs and just make a note about the model naming.

So I agree with @grdryn about uniqueness of the deployment and with @biswassri about not duplicating the sed command all over the pipeline, rather having it in a separate "sanitize" task.

manifests/tasks/test-model-rest-svc-task.yaml
jackdelahunt: Okay I think all of this sounds good. Adding a task for sanitising the name is not what I had in mind first but I think that is a good solution.

jackdelahunt:
@grdryn 

> My first thought after going through this was that there are some of these tasks that we don't actually provide some of these tasks (we just use the openshift-client cluster task directly in the pipeline, rather than having any custom task of our own), so maybe the onus is on customers to provide valid input scripts when using these tasks in their own pipelines?

Just as a side to this PR this was a thought that I have had for a while. The goal of these refactors is to have a task first approach and let the users compose pipeline for their needs.

If we want all of the feature that were a part of the aiedge pipeline (in this case the testing) to be usable in a users pipelines what do we do about the parts that are scripts passed to tasks in the pipeline and not tasks themselves.

grdryn:
> If we want all of the feature that were a part of the aiedge pipeline (in this case the testing) to be usable in a users pipelines what do we do about the parts that are scripts passed to tasks in the pipeline and not tasks themselves.

I'd hesitate to start wrapping those existing generic `openshift-client` tasks or making equivalents for now. They can just add what they need to their pipeline in the same way, right? They could copy from our examples, or deviate for their particular case: maybe they need to use something like a StatefulSet or DaemonSet rather than Deployment like we've got, or they want to deploy (and cleanup) the test resources using GitOps or something. 

I'm not sure if these scenarios are realistic or not, but I'd vote to hold off for now, rather than trying to create new tasks that might need a lot of params. WDYT?

jackdelahunt:
@grdryn Yeah I am not really sure on what the best course of action is and waiting for real world use cases for us to look at will be useful. 

My initial question was raised because I was thinking there is value in those parts of the pipeline where we use the `openshift-client` task and we might want to provide that.

manifests/tasks/object-name-sanitise.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: object-name-sanitise
```
MarianMacik, line 4:
```suggestion
  name: sanitise-object-name
```
reads better maybe?

manifests/tasks/object-name-sanitise.yaml
```diff
@@ -0,0 +1,23 @@
...
+spec:
+  description: This Task can be used to sanitise a string that will be used for naming a k8s object. This task converts the string to a value that can be used for an object.
```
MarianMacik, line 6: Maybe it would be good to include how it is sanitized (forbidden chars replaced with `-`)?

manifests/tasks/object-name-sanitise.yaml
```diff
@@ -0,0 +1,23 @@
...
+metadata:
+  name: object-name-sanitise
+spec:
+  description: This Task can be used to sanitise a string that will be used for naming a k8s object. This task converts the string to a value that can be used for an object. 
+  params:
+  - name: input-string
+    type: string
+  results:
+  - name: output-string
+    description: Total size of the model files
```
MarianMacik, line 12:
```suggestion
    description: Sanitized output
```

manifests/tasks/object-name-sanitise.yaml
```diff
@@ -0,0 +1,23 @@
...
+  steps:
+  - name: sanitise-input-string
+    image: registry.access.redhat.com/ubi9/ubi-minimal
+    script: |
+      #!/usr/bin/env bash
+
+      set -Eeuo pipefail
+
+      echo $(params.input-string) | sed 's/[^a-z0-9.-]/-/g; s/-*$//; s/\n*$//; s/ *$//' | tr -d "\n"  > $(results.output-string.path);
```
MarianMacik, line 21:
```suggestion
      echo $(params.input-string) | sed 's/[^a-z0-9.-]/-/g; s/-*$// | tr -d "\n"  > $(results.output-string.path);
```
replacement of `\n` is done by `tr` and do we need the replacement of empty space if it is replaced by `-` in the first place (because it satisfies `[^a-z0-9.-]`) ?

manifests/tasks/object-name-sanitise.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+    workingDir: /
```
MarianMacik, line 23: For brevity I would even remove this.

manifests/pipelines/aiedge-e2e.pipeline.yaml
```diff
@@ -310,7 +319,7 @@ spec:
   - name: test-model-rest-svc
     params:
     - name: model-name
-      value: $(params.model-name)
+      value: $(tasks.sanitise-model-name.results.output-string)
     - name: model-version
       value: $(params.model-version)
```
MarianMacik, line 324: As @grdryn suggested [here](https://github.com/opendatahub-io/ai-edge/pull/280#pullrequestreview-2215325557) we can parameterize this with the service name rather than model-name and model-version. This will make the task more configurable and users can choose their own deployment and service naming. Now they are basically forced to use `$(tasks.sanitise-model-name.results.output-string)-$(params.model-version)` if they want to use this testing task.

manifests/tasks/object-name-sanitise.yaml
jackdelahunt: Sure

manifests/tasks/object-name-sanitise.yaml
jackdelahunt: Which part the working dir?

manifests/pipelines/aiedge-e2e.pipeline.yaml
jackdelahunt: Yep makes sense will change

manifests/tasks/object-name-sanitise.yaml
MarianMacik: @jackdelahunt oh, my bad, I meant `workingDir: /`

manifests/pipelines/aiedge-e2e.pipeline.yaml
```diff
@@ -309,10 +318,8 @@ spec:
       name: openshift-client
   - name: test-model-rest-svc
     params:
-    - name: model-name
-      value: $(params.model-name)
-    - name: model-version
-      value: $(params.model-version)
+    - name: deployment-name
```
MarianMacik, line 321: I don't want to be nitpicking, it's just this is in fact a name of the service. Even the task is named `test-model-rest-svc`, so maybe we should align it?
```suggestion
    - name: service-name
```

manifests/pipelines/s3-fetch-pipeline.yaml
```diff
@@ -249,10 +258,8 @@ spec:
       name: openshift-client
   - name: test-model-rest-svc
     params:
-    - name: model-name
-      value: $(params.model-name)
-    - name: model-version
-      value: $(params.model-version)
+    - name: deployment-name
```
MarianMacik, line 261:
```suggestion
    - name: service-name
```

manifests/tasks/test-model-rest-svc-task.yaml
```diff
@@ -4,10 +4,7 @@ metadata:
   name: test-model-rest-svc
 spec:
   params:
-  - name: model-name
-    type: string
-  - name: model-version
-    default: "1"
+  - name: deployment-name
```
MarianMacik, line 7:
```suggestion
  - name: service-name
```

manifests/tasks/test-model-rest-svc-task.yaml
```diff
@@ -26,7 +23,7 @@ spec:
         echo "\nCall service:"
 
         #  call service
-        curl -v -X POST -H "Content-Type:application/json" --data @$(workspaces.test-data.path)/data.json -o /tmp/output.json http://$(params.model-name)-$(params.model-version):8080/$(params.test-endpoint)
+        curl -v -X POST -H "Content-Type:application/json" --data @$(workspaces.test-data.path)/data.json -o /tmp/output.json http://$(params.deployment-name):8080/$(params.test-endpoint)
```
MarianMacik, line 26:
```suggestion
        curl -v -X POST -H "Content-Type:application/json" --data @$(workspaces.test-data.path)/data.json -o /tmp/output.json http://$(params.service-name):8080/$(params.test-endpoint)
```

manifests/tasks/object-name-sanitise.yaml
jackdelahunt: No problem

manifests/tasks/test-model-rest-svc-task.yaml
jackdelahunt: Done

jackdelahunt:
@MarianMacik all comments addressed

MarianMacik (approved):
Looks good to me, nice!
/approve

manifests/tasks/sanitise-object-name.yaml
```diff
@@ -0,0 +1,21 @@
...
+    description: Sanitised output string
+  steps:
+  - name: sanitise-input-string
+    image: registry.access.redhat.com/ubi9/ubi-minimal
+    script: |
+      #!/usr/bin/env bash
+
+      set -Eeuo pipefail
+
+      echo $(params.input-string) | sed 's/[^a-z0-9.-]/-/g; s/-*$//' | tr -d "\n"  > $(results.output-string.path);
```
grdryn, line 21: There are some other possible checks we might want to do in future to the proposed name before returning, now that this is a generic task dedicated to that purpose.

E.g. check length and shorten if too long (or maybe error if empty?)

grdryn (approved):
/lgtm :shipit:

manifests/tasks/sanitise-object-name.yaml
grdryn: Also the end currently just checks to make sure it's not a `-`, but I guess it also needs to check that it's not a `.`, right? Similarly the start of the string should be checked at that point for both characters, since they are invalid starting characters, and even though the input string may not have begun with them, we may have dropped other invalid characters preceding them in the first sed expression, right?

Also, maybe the expected behaviour for A-Z chars is that they're translated to a-z, rather than dropped.
E.g.:
- MY-MODEL-123 -> my-model-123 (rather than just 123)
- My-MoDeL-123 -> my-model-123 (rather than y-oe-123)

Thoughts?

jackdelahunt: Yeah I think that all makes sense the point of checking if `-` is at the start is a good catch because that isn't valid either.

And I think change the A-Z to lowercase is a good idea I will try and figure that part out now

jackdelahunt:
@grdryn what do you think now

- I separated each part out for easier readability
- I added the replacing of uppercase to lowercase
- I am now trimming from both ends

grdryn (approved):
/lgtm

Thanks @jackdelahunt!

PR #283, opened 2024-07-31
RHOAIENG-6634: Test stage can use wrong image, because of tag reference and imagePullPolicy

jackdelahunt:
**changes only applied to aiedge pipeline so far**

## Description
In the test stage of the aiedge-e2e pipeline, we deploy the built image by referring to it by the tag reference (rather than manifest digest reference), and we don't explicitly set the imagePullPolicy to Always. This can lead to difficult to a case where an older candidate image is already stored on the node from a previous run, and won't be refreshed before deploying. This could mean that the new image doesn't actually get tested.

## How Has This Been Tested?
- CI
- Run e2e tests as normal and all should pass

## Merge criteria:

### DIFF
--- a/manifests/pipelines/s3-fetch-pipeline.yaml
+++ b/manifests/pipelines/s3-fetch-pipeline.yaml
@@ -204,7 +204,7 @@ spec:
                 app: $(params.model-name)-$(params.model-version)
             spec:
               containers:
-              - image: $(params.candidate-image-tag-reference)
+              - image: $(params.candidate-image-tag-reference)@$(tasks.build-container-image.results.IMAGE_DIGEST)
                 name: $(params.model-name)-$(params.model-version)
                 livenessProbe:
                   failureThreshold: 10
--- a/manifests/pipelines/aiedge-e2e.pipeline.yaml
+++ b/manifests/pipelines/aiedge-e2e.pipeline.yaml
@@ -264,7 +264,7 @@ spec:
                 app: $(params.model-name)-$(params.model-version)
             spec:
               containers:
-              - image: $(params.candidate-image-tag-reference)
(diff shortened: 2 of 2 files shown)


manifests/pipelines/s3-fetch-pipeline.yaml
```diff
@@ -204,7 +204,7 @@ spec:
                 app: $(params.model-name)-$(params.model-version)
             spec:
               containers:
-              - image: $(params.candidate-image-tag-reference)
+              - image: $(params.candidate-image-tag-reference)@$(tasks.build-container-image.results.IMAGE_DIGEST)
```
MarianMacik, line 207: Shouldn't the imagePullPolicy be added here as well?

```suggestion
              - image: $(params.candidate-image-tag-reference)@$(tasks.build-container-image.results.IMAGE_DIGEST)
                imagePullPolicy: Always
```

MarianMacik (commented):
One optimization which I realized when [consulted the docs](https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting) is that if we specify a digest, `imagePullPolicy` will be automatically set to `IfNotPresent`, so in case the digest is the same, it doesn't query the remote registry. With `Always` it would always reach to external registry to check. So maybe just digest is enough, not needing to put `Always` explicitly.

manifests/pipelines/s3-fetch-pipeline.yaml
jackdelahunt: Yep you are right, will add now

MarianMacik (approved):
Looks good now, thanks!
/approve

biswassri:
/lgtm 

Looks good to me too. I'll add this change to the git pipeline as well.

biswassri (approved):

PR #288, opened 2024-08-16
RHOAIENG-205-Define the standard terms

biswassri:
JIRA: [RHOAIENG-205](https://issues.redhat.com/browse/RHOAIENG-205)
## Description

Adding a glossary for stand terms. This is just the first set. We plan to iteratively update this terminology list. Ref: https://github.com/LaVLaS/ai-edge/blob/glossary/docs/glossary.md

I did move the doc that Landon had started but do we want to add more terminology to this? 

## How Has This Been Tested?

N/A

## Merge criteria:

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -12,7 +12,7 @@
 
 ## Model Server
 
-- A Model Server is responsible for hosting models as a service to "to return predictions based on data inputs that you provide through API calls."[^2]
+- A Model Server is responsible for hosting models as a service "to return predictions based on data inputs that you provide through API calls."[^2]
 - For any workflows under opendatahub-io/ai-edge, we will be focusing on using the Model Servers and serving runtimes supported by Open Data Hub
 
 ## Inference Application Container
@@ -30,7 +30,7 @@
 
 ## GitOps
 
-- GitOps is an established configuration management pattern to store the configuration of your infrastructure configuration and workflow automation in a Git repository for reproducibility and version control.
+- GitOps is an established configuration management pattern to store the configuration of your infrastructure and workflow automation in a Git repository for reproducibility and version control.
 - "GitOps uses Git repositories as a single source of truth to deliver infrastructure as code."[^1]
 
 [^1]: [What is GitOps](https://www.redhat.com/en/topics/devops/what-is-gitops)
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -8,25 +8,25 @@
 ## Near Edge
 
 - This is a non-core distributed environment to run and serve AI/ML inference workloads in moderate yet constrained compute resources and network.
-- For the purpose of this repository, the near edge environment is represented by separate OpenShift clusters disconnected from the core, the internet or both but may be managed from a core OpenShift cluster.
+- For the purpose of this repository, the near edge environment is represented by separate OpenShift clusters which may be disconnected from the core, the internet or both but may be managed from a core OpenShift cluster.
 
 ## Model Server
 
 - A Model Server is responsible for hosting models as a service "to return predictions based on data inputs that you provide through API calls."[^2]
-- For any workflows under opendatahub-io/ai-edge, we will be focusing on using the Model Servers and serving runtimes supported by Open Data Hub
+- For any workflows under opendatahub-io/ai-edge, we will be focusing on using the Model Servers and serving runtimes supported by Open Data Hub.
 
 ## Inference Application Container
 
-- OCI compliant container image[^3] with the models included during the build process
-- Support for container images where the model and model serving runtimes are stored together
+- OCI compliant container image[^3] with the models included during the build process.
+- Support for container images where the model and model serving runtimes are stored together.
 
 ## Model Registry
 
-- A centralized repository for the models and their metadata and managing the model lifecycle and versions. Currently our pipelines only support S3 and Git as the source where models can be stored. 
+- A centralized repository for the models and their metadata and managing the model lifecycle and versions. Currently our pipelines do not support any Model Registry; only S3 and Git can be used as directly referenced sources where models can be stored.
 
 ## OCI Distribution Registry
 
-- Open Container Initiative (OCI) compliant container registry where the model and other artifacts are stored and versioned ready to be deployed on production or staging environments.
+- Open Container Initiative (OCI) compliant container registry where the Inference Application Container images and other artifacts are stored and versioned ready to be deployed on production or staging environments.
 
 ## GitOps
 
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -1,0 +1,38 @@
+# AI Edge Terminology
+
+## Core
+
+- The central OpenShift cluster containing the tools responsible for creating any artifacts required for the successful deployment of an Inference Application to Near Edge environments.
+- There are no resources or network constraints expected in the core cluster as it is expected that it fully supports all workflows required for creating and verifying Inference Application container images.
+
+## Near Edge
+
+- This is a non-core distributed environment to run and serve AI/ML inference workloads in moderate yet constrained compute resources and network.
+- For the purpose of this repository, the near edge environment is represented by separate OpenShift clusters disconnected from the core, the internet or both but may be managed from a core OpenShift cluster.
+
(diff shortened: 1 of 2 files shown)


jackdelahunt (approved):
/approve

docs/glossary.md
```diff
@@ -0,0 +1,38 @@
...
+- There are no resources or network constraints expected in the core cluster as it is expected that it fully supports all workflows required for creating and verifying Inference Application container images.
+
+## Near Edge
+
+- This is a non-core distributed environment to run and serve AI/ML inference workloads in moderate yet constrained compute resources and network.
+- For the purpose of this repository, the near edge environment is represented by separate OpenShift clusters disconnected from the core, the internet or both but may be managed from a core OpenShift cluster.
+
+## Model Server
+
+- A Model Server is responsible for hosting models as a service to "to return predictions based on data inputs that you provide through API calls."[^2]
```
MarianMacik, line 15:
```suggestion
- A Model Server is responsible for hosting models as a service "to return predictions based on data inputs that you provide through API calls."[^2]
```

docs/glossary.md
```diff
@@ -0,0 +1,38 @@
...
+
+- A centralized repository for the models and their metadata and managing the model lifecycle and versions. Currently our pipelines only support S3 and Git as the source where models can be stored. 
+
+## OCI Distribution Registry
+
+- Open Container Initiative (OCI) compliant container registry where the model and other artifacts are stored and versioned ready to be deployed on production or staging environments.
+
+## GitOps
+
+- GitOps is an established configuration management pattern to store the configuration of your infrastructure configuration and workflow automation in a Git repository for reproducibility and version control.
```
MarianMacik, line 33:
```suggestion
- GitOps is an established configuration management pattern to store the configuration of your infrastructure and workflow automation in a Git repository for reproducibility and version control.
```

MarianMacik:
Rebase is needed

docs/glossary.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Near Edge
+
+- This is a non-core distributed environment to run and serve AI/ML inference workloads in moderate yet constrained compute resources and network.
+- For the purpose of this repository, the near edge environment is represented by separate OpenShift clusters disconnected from the core, the internet or both but may be managed from a core OpenShift cluster.
+
+## Model Server
+
+- A Model Server is responsible for hosting models as a service "to return predictions based on data inputs that you provide through API calls."[^2]
+- For any workflows under opendatahub-io/ai-edge, we will be focusing on using the Model Servers and serving runtimes supported by Open Data Hub
```
StevenTobin, line 16:
```suggestion
- For any workflows under opendatahub-io/ai-edge, we will be focusing on using the Model Servers and serving runtimes supported by Open Data Hub.
```

docs/glossary.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Inference Application Container
+
+- OCI compliant container image[^3] with the models included during the build process
```
StevenTobin, line 20:
```suggestion
- OCI compliant container image[^3] with the models included during the build process.
```

docs/glossary.md
```diff
@@ -0,0 +1,38 @@
...
+- Support for container images where the model and model serving runtimes are stored together
```
StevenTobin, line 21:
```suggestion
- Support for container images where the model and model serving runtimes are stored together.
```

StevenTobin (commented):
Tiny nit-pick.

MarianMacik (approved):
Looks good to me!

docs/glossary.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Core
+
+- The central OpenShift cluster containing the tools responsible for creating any artifacts required for the successful deployment of an Inference Application to Near Edge environments.
+- There are no resources or network constraints expected in the core cluster as it is expected that it fully supports all workflows required for creating and verifying Inference Application container images.
+
+## Near Edge
+
+- This is a non-core distributed environment to run and serve AI/ML inference workloads in moderate yet constrained compute resources and network.
+- For the purpose of this repository, the near edge environment is represented by separate OpenShift clusters disconnected from the core, the internet or both but may be managed from a core OpenShift cluster.
```
adelton, line 11: This "disconnected from the core" -- is that a requirement (no connection must be possible from the core to the near edge clusters (or vice versa?)), or a possible setup (even if no connections are possible from the core to the near edge clusters (or vice versa?) the setup still applies)?

docs/glossary.md
```diff
@@ -0,0 +1,38 @@
...
+- For any workflows under opendatahub-io/ai-edge, we will be focusing on using the Model Servers and serving runtimes supported by Open Data Hub
+
+## Inference Application Container
+
+- OCI compliant container image[^3] with the models included during the build process
+- Support for container images where the model and model serving runtimes are stored together
+
+## Model Registry
+
+- A centralized repository for the models and their metadata and managing the model lifecycle and versions. Currently our pipelines only support S3 and Git as the source where models can be stored.
```
adelton, line 25: My understanding is that Model Registry would store only the model metadata, not the actual models anyway.

So should we clarify that currently our pipelines do not support any Model Registry and locations of models (either in S3 or in git repos) need to be specified directly?

docs/glossary.md
```diff
@@ -0,0 +1,38 @@
...
+- OCI compliant container image[^3] with the models included during the build process
+- Support for container images where the model and model serving runtimes are stored together
+
+## Model Registry
+
+- A centralized repository for the models and their metadata and managing the model lifecycle and versions. Currently our pipelines only support S3 and Git as the source where models can be stored. 
+
+## OCI Distribution Registry
+
+- Open Container Initiative (OCI) compliant container registry where the model and other artifacts are stored and versioned ready to be deployed on production or staging environments.
```
adelton, line 29: Here we assume that the model will be in the registry ... but above we say that we support models only in S3 or git repository.

So are we currently able to use container registries as sources of models?

docs/glossary.md
biswassri: So, in this terminology we are trying to define Model Registry as a generic model registry and not the ODH Model Registry. 

> So should we clarify that currently our pipelines do not support any Model Registry and locations of models (either in S3 or in git repos) need to be specified directly?

^This I think is clarified by the last line. `Currently our pipelines only support S3 and Git as the source where models can be stored. ` Since we are trying to keep the terminology generic, we didn't want to mention that we don't support any other registry. Does that make sense?

docs/glossary.md
biswassri: It's actually a possible scenerio when describing the `near-edge`, the disconnected environment.  From the original architecture the pipelines were supposed to be running in a cluster that is connected to the Internet. 

@MarianMacik suggests the below could possibly be a better wording of the near-edge : 

`For the purpose of this repository, the near edge environment is represented by separate OpenShift clusters which _may be_ disconnected from the core, the internet or both but may be managed from a core OpenShift cluster.`

Wdyt?

docs/glossary.md
biswassri: My thought was that quay is an OCI compliant registry and thats where our model containers are stored at the end of the pipeline, and hence this definition. We are only supporting s3 and git as the source of the models. Does that make sense?

adelton: Right, but then we should say "where the Inference Application Container images are stored ...", not "where the model and other artifacts are stored ..." because the current wording suggests that we are ready to support something more being fetched from the image registry (for example models during the build pipeline) than we are.

biswassri: Makes sense, I can update the definition to : 

`Open Container Initiative (OCI) compliant container registry where the Inference Application Container images and other artifacts are stored and versioned ready to be deployed on production or staging environments.`

biswassri:
@StevenTobin @adelton I have made the changes according to your comments, if everything looks good can any of you mark it approved?

StevenTobin:
lgtm but @adelton's comments were more substantive than mine so I'll let him approve.

adelton:
LGTM.

adelton (approved):

jackdelahunt (approved):
/approve

PR #290, opened 2024-08-19
RHOAIENG-209-Deprecate the use of Tekton ClusterTasks in any Pipelines

biswassri:
JIRA: [RHOAIENG-209](https://issues.redhat.com/browse/RHOAIENG-209)
## Description

Deprecating clustertasks and replacing with git-resolvers. The idea is we can currently pulling in pre-defined task yamls from the tekton catalog git repo. Ref: [Migration from clustertasks to tekton-resolvers ](https://www.redhat.com/en/blog/migration-from-clustertasks-to-tekton-resolvers-in-openshift-pipelines)

## How Has This Been Tested?

```
oc apply -k manifests/
oc create -f examples/tekton/aiedge-e2e/example-pipelineruns/git-fetch.tensorflow-housing.pipelinerun.yaml
```
## Merge criteria:

### DIFF
--- a/manifests/pipelines/git-fetch-pipeline.yaml
+++ b/manifests/pipelines/git-fetch-pipeline.yaml
@@ -382,14 +382,12 @@ spec:
   - name: skopeo-copy
     params:
     - name: SOURCE_IMAGE_URL
-      value: $(tasks.retrieve-build-image-info.source_image_url.path)
+      value: $(tasks.retrieve-build-image-info.source-image-url)
     - name: DESTINATION_IMAGE_URL
-      value: $(tasks.retrieve-build-image-info.destination_image_url.path)
+      value: $(tasks.retrieve-build-image-info.dest-image-url)
     - name: SRC_TLS_VERIFY
       value: "true"
     - name: DEST_TLS_VERIFY
-      value: "true"
-    - name: VERBOSE
       value: "true"
     runAfter:
     - retrieve-build-image-info
--- a/manifests/pipelines/git-fetch-pipeline.yaml
+++ b/manifests/pipelines/git-fetch-pipeline.yaml
@@ -287,11 +287,11 @@ spec:
       resolver: git
       params:
         - name: url
-          value: https://github.com/tektoncd/catalog.git
+          value: https://github.com/openshift-pipelines/tektoncd-catalog
         - name: revision
-          value: main
+          value: p
         - name: pathInRepo
-          value: task/openshift-client/0.2/openshift-client.yaml
+          value: tasks/task-openshift-client/0.2.1/task-openshift-client.yaml
     runAfter:
       - build-container-image
   - name: create-default-service
@@ -308,11 +308,11 @@ spec:
       resolver: git
       params:
         - name: url
-          value: https://github.com/tektoncd/catalog.git
+          value: https://github.com/openshift-pipelines/tektoncd-catalog
         - name: revision
-          value: main
+          value: p
         - name: pathInRepo
-          value: task/openshift-client/0.2/openshift-client.yaml
+          value: tasks/task-openshift-client/0.2.1/task-openshift-client.yaml
   - name: test-model-rest-svc
     params:
     - name: service-name
@@ -349,11 +349,11 @@ spec:
       resolver: git
       params:
         - name: url
-          value: https://github.com/tektoncd/catalog.git
+          value: https://github.com/openshift-pipelines/tektoncd-catalog
         - name: revision
-          value: main
+          value: p
         - name: pathInRepo
-          value: task/openshift-client/0.2/openshift-client.yaml
+          value: tasks/task-openshift-client/0.2.1/task-openshift-client.yaml
   - name: retrieve-build-image-info
     taskRef:
       kind: Task
@@ -382,12 +382,12 @@ spec:
   - name: skopeo-copy
     params:
     - name: SOURCE_IMAGE_URL
-      value: $(tasks.retrieve-build-image-info.source-image-url)
+      value: $(tasks.retrieve-build-image-info.results.source-image-url)
     - name: DESTINATION_IMAGE_URL
-      value: $(tasks.retrieve-build-image-info.dest-image-url)
-    - name: SRC_TLS_VERIFY
+      value: $(tasks.retrieve-build-image-info.results.dest-image-url)
+    - name: srcTLSverify
       value: "true"
-    - name: DEST_TLS_VERIFY
+    - name: destTLSverify
       value: "true"
     runAfter:
     - retrieve-build-image-info
--- a/manifests/pipelines/git-fetch-pipeline.yaml
+++ b/manifests/pipelines/git-fetch-pipeline.yaml
@@ -377,7 +377,7 @@ spec:
     - test-model-rest-svc
     - stop-deployment
     workspaces:
-    - name: images-url
+    - name: images_url
       workspace: build-workspace-pv
   - name: skopeo-copy
     params:
@@ -385,6 +385,10 @@ spec:
       value: ""
     - name: DESTINATION_IMAGE_URL
       value: ""
+    - name: SRC_TLS_VERIFY
+      value: "true"
+    - name: DEST_TLS_VERIFY
+      value: "true"
     runAfter:
     - retrieve-build-image-info
     taskRef:
@@ -397,7 +401,7 @@ spec:
         - name: pathInRepo
           value: tasks/task-skopeo-copy/0.4.2/task-skopeo-copy.yaml
     workspaces:
-    - name: images-url
+    - name: images_url
       workspace: build-workspace-pv
   workspaces:
   - name: build-workspace-pv
--- a/manifests/pipelines/git-fetch-pipeline.yaml
+++ b/manifests/pipelines/git-fetch-pipeline.yaml
@@ -381,15 +381,21 @@ spec:
       workspace: build-workspace-pv
   - name: skopeo-copy
     params:
-    - name: srcTLSverify
-      value: "true"
-    - name: destTLSverify
-      value: "true"
+    - name: SOURCE_IMAGE_URL
+      value: ""
+    - name: DESTINATION_IMAGE_URL
+      value: ""
     runAfter:
     - retrieve-build-image-info
     taskRef:
-      kind: ClusterTask
-      name: skopeo-copy
+      resolver: git
+      params:
+        - name: url
+          value: https://github.com/openshift-pipelines/tektoncd-catalog
+        - name: revision
+          value: p
+        - name: pathInRepo
+          value: tasks/task-skopeo-copy/0.4.2/task-skopeo-copy.yaml
     workspaces:
     - name: images-url
       workspace: build-workspace-pv
--- a/manifests/pipelines/git-fetch-pipeline.yaml
+++ b/manifests/pipelines/git-fetch-pipeline.yaml
@@ -287,11 +287,11 @@ spec:
       resolver: git
       params:
         - name: url
-          value: https://github.com/tektoncd/catalog.git
+          value: https://github.com/openshift-pipelines/tektoncd-catalog
         - name: revision
-          value: main
+          value: p
         - name: pathInRepo
-          value: task/openshift-client/0.2/openshift-client.yaml
+          value: tasks/task-openshift-client/0.2.0/task-openshift-client.yaml
     runAfter:
       - build-container-image
   - name: create-default-service
@@ -308,11 +308,11 @@ spec:
       resolver: git
       params:
         - name: url
-          value: https://github.com/tektoncd/catalog.git
+          value: https://github.com/openshift-pipelines/tektoncd-catalog
         - name: revision
-          value: main
+          value: p
         - name: pathInRepo
-          value: task/openshift-client/0.2/openshift-client.yaml
+          value: tasks/task-openshift-client/0.2.0/task-openshift-client.yaml
   - name: test-model-rest-svc
     params:
     - name: service-name
@@ -349,11 +349,11 @@ spec:
       resolver: git
       params:
         - name: url
-          value: https://github.com/tektoncd/catalog.git
+          value: https://github.com/openshift-pipelines/tektoncd-catalog
         - name: revision
-          value: main
+          value: p
         - name: pathInRepo
-          value: task/openshift-client/0.2/openshift-client.yaml
+          value: tasks/task-openshift-client/0.2.0/task-openshift-client.yaml
   - name: retrieve-build-image-info
     taskRef:
       kind: Task
@@ -377,17 +377,19 @@ spec:
     - test-model-rest-svc
     - stop-deployment
     workspaces:
-    - name: images-url
+    - name: images_url
       workspace: build-workspace-pv
   - name: skopeo-copy
     params:
     - name: SOURCE_IMAGE_URL
-      value: docker://docker.io/busybox:latest
+      value: $(tasks.retrieve-build-image-info.source_image_url.path)
     - name: DESTINATION_IMAGE_URL
-      value: docker://image-registry.openshift-image-registry.svc:5000/task-containers/busybox:latest
-    - name: srcTLSverify
+      value: $(tasks.retrieve-build-image-info.destination_image_url.path)
+    - name: SRC_TLS_VERIFY
       value: "true"
-    - name: destTLSverify
+    - name: DEST_TLS_VERIFY
+      value: "true"
+    - name: VERBOSE
       value: "true"
     runAfter:
     - retrieve-build-image-info
@@ -401,7 +403,7 @@ spec:
         - name: namespace
           value: openshift-pipelines
     workspaces:
-    - name: images-url
+    - name: images_url
       workspace: build-workspace-pv
   workspaces:
   - name: build-workspace-pv
--- a/manifests/pipelines/git-fetch-pipeline.yaml
+++ b/manifests/pipelines/git-fetch-pipeline.yaml
@@ -108,8 +108,14 @@ spec:
   tasks:
   - name: fetch-model-git
     taskRef:
-      kind: ClusterTask
-      name: git-clone
+      resolver: git
+      params:
+        - name: url
+          value: https://github.com/tektoncd/catalog.git
+        - name: revision
+          value: main
+        - name: pathInRepo
+          value: task/git-clone/0.9/git-clone.yaml
     params:
     - name: url
       value: $(params.git-model-repo)
@@ -141,6 +147,15 @@ spec:
     - name: workspace
       workspace: build-workspace-pv
   - name: git-clone-containerfile-repo
+    taskRef:
+      resolver: git
+      params:
+        - name: url
+          value: https://github.com/tektoncd/catalog.git
+        - name: revision
+          value: main
+        - name: pathInRepo
+          value: task/git-clone/0.9/git-clone.yaml  
     params:
     - name: url
       value: $(params.git-containerfile-repo)
@@ -150,9 +165,6 @@ spec:
       value: /containerfile_repo/
     runAfter:
     - move-model-to-root-dir
-    taskRef:
-      kind: ClusterTask
-      name: git-clone
     workspaces:
     - name: output
       workspace: build-workspace-pv
@@ -264,8 +276,14 @@ spec:
         oc wait deployment -n $(params.target-namespace) $(tasks.sanitise-model-name.results.output-string)-$(params.model-version) --for condition=Available=True --timeout=120s
         oc wait pod -n $(params.target-namespace) -l app=$(params.model-name)-$(params.model-version) --for condition=Ready=True --timeout=120s
     taskRef:
-      kind: ClusterTask
-      name: openshift-client
+      resolver: git
+      params:
+        - name: url
+          value: https://github.com/tektoncd/catalog.git
+        - name: revision
+          value: main
+        - name: pathInRepo
+          value: task/openshift-client/0.2/openshift-client.yaml
     runAfter:
       - build-container-image
   - name: create-default-service
@@ -279,8 +297,14 @@ spec:
     runAfter:
     - test-container-deploy
     taskRef:
-      kind: ClusterTask
-      name: openshift-client
+      resolver: git
+      params:
+        - name: url
+          value: https://github.com/tektoncd/catalog.git
+        - name: revision
+          value: main
+        - name: pathInRepo
+          value: task/openshift-client/0.2/openshift-client.yaml
   - name: test-model-rest-svc
     params:
     - name: service-name
@@ -314,8 +338,14 @@ spec:
     runAfter:
     - test-model-rest-svc
     taskRef:
-      kind: ClusterTask
-      name: openshift-client
+      resolver: git
+      params:
+        - name: url
+          value: https://github.com/tektoncd/catalog.git
+        - name: revision
+          value: main
+        - name: pathInRepo
+          value: task/openshift-client/0.2/openshift-client.yaml
   - name: retrieve-build-image-info
     taskRef:
       kind: Task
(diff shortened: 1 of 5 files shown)


biswassri:
The change for the `build-container-image` is still something I'm figuring out. Using the same method as others, I'm getting errors. So still working on it.

MarianMacik:
Rebase is needed

manifests/pipelines/git-fetch-pipeline.yaml
```diff
@@ -348,8 +378,14 @@ spec:
...
-      kind: ClusterTask
-      name: skopeo-copy
+      resolver: git
+      params:
+        - name: url
+          value: https://github.com/tektoncd/catalog.git
+        - name: revision
+          value: main
+        - name: pathInRepo
+          value: task/skopeo-copy/0.3/skopeo-copy.yaml
```
MarianMacik, line 388: I noticed that now it uses images from `quay.io`. Before it used cluster tasks installed by OpenShift Pipelines, which means it used images from `registry.redhat.io`. This may be the reason why `skopeo-copy` task is failing with:
```
level=fatal msg="initializing source docker://image-registry.openshift-image-registry.svc:5000/pipeline-test-tensorflow-housing/tensorflow-housing@sha256:2b6b85d6c3998ae2efb73c931755710bb7ac000925bc6c15479b6341ac3d1bff: reading manifest sha256:2b6b85d6c3998ae2efb73c931755710bb7ac000925bc6c15479b6341ac3d1bff in image-registry.openshift-image-registry.svc:5000/pipeline-test-tensorflow-housing/tensorflow-housing: authentication required"
```
So maybe with Tekton task the authentication is not working correctly with OpenShift Pipelines operator and there may be some incompatibility. I have rerun the PR check to see if it is persistent. We may need to use the tasks directly from RH to ensure the compatibility.

manifests/pipelines/git-fetch-pipeline.yaml
```diff
@@ -108,8 +108,14 @@ spec:
   tasks:
   - name: fetch-model-git
```
MarianMacik, line 109: S3 version of the pipeline and GitOps pipeline should also have these, right?

biswassri: Yeap, I have that ready. Just haven't pushed it because of the buildah one.

manifests/pipelines/s3-fetch-pipeline.yaml
```diff
@@ -289,8 +315,14 @@ spec:
...
-      kind: ClusterTask
-      name: openshift-client
+      resolver: git
+      params:
+        - name: url
+          value: https://github.com/tektoncd/catalog.git
+        - name: revision
+          value: main
+        - name: pathInRepo
+          value: task/openshift-client/0.2/openshift-client.yaml
```
biswassri, line 325: @MarianMacik I'll update this once this PR is merged in https://github.com/openshift-pipelines/task-openshift/pull/47

MarianMacik:
There is an error `'[User error] workspace binding "images-url" does not match any declared workspace'` on the `skopeo-copy` task. Which is weird, given the previous `retrieve-build-image-info` task with exactly the same workspace binding passed. Probably an issue on the cluster with Persistence volumes?

MarianMacik:
@biswassri I understand now. The message `'[User error] workspace binding "images-url" does not match any declared workspace'` means that our binding named images-url doesn't match any workspace which is declared on that task, in this case `skopeo-copy`.
I see that we use the provided task by the Pipelines operator and based on this docs: https://docs.openshift.com/pipelines/1.15/create/remote-pipelines-tasks-resolvers.html#op-taskref-skopeo-copy_remote-pipelines-tasks-resolvers
There were changes to the original `ClusterTask`:

Changes from the skopeo-copy ClusterTask:

- All parameter names were changed to uppercase.
- The VERBOSE parameter was added.
- The workspace name was changed from images-url to images_url.
- The SOURCE_DIGEST and DESTINATION_DIGEST results were added.

So we need do change that. Ideally unify across whole pipeline, this means change also the definition of the `retrieve-build-image-info` task we ship.

manifests/pipelines/git-fetch-pipeline.yaml
```diff
@@ -339,21 +377,33 @@ spec:
...
   - name: skopeo-copy
     params:
-    - name: srcTLSverify
+    - name: SOURCE_IMAGE_URL
+      value: $(tasks.retrieve-build-image-info.source_image_url.path)
+    - name: DESTINATION_IMAGE_URL
+      value: $(tasks.retrieve-build-image-info.destination_image_url.path)
+    - name: SRC_TLS_VERIFY
+      value: "true"
+    - name: DEST_TLS_VERIFY
```
MarianMacik, line 390: Pipeline is failing because of this. We should use the `images_url` workspace as before. The README of that Tekton task is missing the information about workspaces but the `images_url` workspace is still there and can be used as before, see [here](https://github.com/openshift-pipelines/tektoncd-catalog/blob/5a1664f1adf1f6cc1323c6cc2a2bdf390658defb/tasks/task-skopeo-copy/0.4.1/task-skopeo-copy.yaml#L30-L33).

manifests/pipelines/git-fetch-pipeline.yaml
```diff
@@ -339,21 +377,33 @@ spec:
...
       value: "true"
-    - name: destTLSverify
+    - name: VERBOSE
       value: "true"
```
MarianMacik, line 393: Also this VERBOSE config is not needed I think.

manifests/pipelines/git-fetch-pipeline.yaml
biswassri: @MarianMacik  I think the pipeline is failing because of this: `level=fatal msg="Invalid source name $(tasks.retrieve-build-image-info.source_image_url.path): Invalid image name \"$(tasks.retrieve-build-image-info.source_image_url.path)\", expected colon-separated transport:reference"` I updated the retrieve-build-image-info task and I think that change is causing the above error and the pipeline failure.

MarianMacik: @biswassri But that is the thing, we shoudn't change the way we specify images in the retrieve-build-image-info task. That should stay the same as it was before, that is using the url.txt file. Or is there any reason for this change?

biswassri: Yes, If you see the new skopeo-copy script, they are no longer user URL.txt file. The new task has added params for source and destination URLs. [here](https://github.com/openshift-pipelines/tektoncd-catalog/blob/5a1664f1adf1f6cc1323c6cc2a2bdf390658defb/tasks/task-skopeo-copy/0.4.1/task-skopeo-copy.yaml#L35-L43)

biswassri:
@MarianMacik @jackdelahunt This is ready for review. Please /lgtm whenever ready

manifests/pipelines/git-fetch-pipeline.yaml
MarianMacik: @biswassri It still has it. [This is the latest version 0.4.1](https://github.com/openshift-pipelines/tektoncd-catalog/blob/p/tasks/task-skopeo-copy/0.4.1/task-skopeo-copy.yaml#L106). It also still has the [workspace definition and mention of copying multiple images](https://github.com/openshift-pipelines/tektoncd-catalog/blob/p/tasks/task-skopeo-copy/0.4.1/task-skopeo-copy.yaml#L29-L33).

When I decode the linked skopeo-copy.sh script, I get this:

```bash
#!/usr/bin/env bash

shopt -s inherit_errexit
set -eu -o pipefail

source "$(dirname "${BASH_SOURCE[0]}")/common.sh"
source "$(dirname "${BASH_SOURCE[0]}")/skopeo-common.sh"

# Ensure the /tekton/home/.docker directory exists
mkdir -p /workspace/home/.docker

phase "Copying '${PARAMS_SOURCE_IMAGE_URL}' into '${PARAMS_DESTINATION_IMAGE_URL}'"

set -x

if [ -n "${PARAMS_SOURCE_IMAGE_URL}" ] && [ -n "${PARAMS_DESTINATION_IMAGE_URL}" ]; then
    skopeo copy ${SKOPEO_DEBUG_FLAG} \
        --src-tls-verify="${PARAMS_SRC_TLS_VERIFY}" \
        --dest-tls-verify="${PARAMS_DEST_TLS_VERIFY}" \
        "${PARAMS_SOURCE_IMAGE_URL}" \
        "${PARAMS_DESTINATION_IMAGE_URL}"
else
    # Function to copy multiple images.
    copyimages() {
        filename="${WORKSPACES_IMAGES_URL_PATH}/url.txt"
        while IFS= read -r line || [ -n "$line" ]
        do
            cmd=""
            for url in $line
            do
                cmd="$cmd $url"
            done
            read -ra SOURCE <<<"${cmd}"
            skopeo copy "${SOURCE[@]}" --src-tls-verify="${PARAMS_SRC_TLS_VERIFY}" --dest-tls-verify="${PARAMS_DEST_TLS_VERIFY}"
            echo "$cmd"
        done < "$filename"
    }

    copyimages
fi
```
which clearly shows the possibility to use the `url.txt` file.

It is available since version 0.3.1.

biswassri: @MarianMacik okay sorry, I actually hadn't decoded the skopeo-copy script. I'll revert the retrieve info script changes. But in that case, what would we populate `SOURCE_IMAGE_URL` and `DESTINATION_IMAGE_URL` with? I thought that this was to be populated as a task result from retrieve info.

MarianMacik: @biswassri I see. In that case they should be empty. Otherwise the url.txt file won't be taken into consideration.

manifests/tasks/retrieve-build-image-info/retrieve-build-image-info.yaml
```diff
@@ -26,9 +29,6 @@ spec:
   - name: target-image-tag-references
     type: array
     description: The image tag references used for the final built image
-  workspaces:
-  - name: images-url
-    description: workspace where url.txt file is created
```
MarianMacik, line 31: I think in this task we just need to change the workspace name to `images_url` and that's all.

MarianMacik (approved):

manifests/tasks/retrieve-build-image-info/retrieve-build-image-info.yaml
MarianMacik, line 31: I think here we just need to change this workspace name to `images_url` and that's all.

manifests/pipelines/git-fetch-pipeline.yaml
```diff
@@ -279,8 +305,14 @@ spec:
...
     - test-container-deploy
     taskRef:
-      kind: ClusterTask
-      name: openshift-client
+      resolver: git
+      params:
+        - name: url
+          value: https://github.com/openshift-pipelines/tektoncd-catalog
+        - name: revision
+          value: p
```
jackdelahunt, line 313: I was going to say this is wrong but I checked and it is actually called p...

jackdelahunt (changes requested):
This panic is getting triggered in the e2e-tests because there is a new possible state for the task https://github.com/opendatahub-io/ai-edge/blob/main/test/e2e-tests/tests/pipelines_test.go#L296-L309

Adding a new case called "ResolvingTaskRef" that returns `false, nil` should fix the issue

jackdelahunt:
I am also getting other errors in the e2e-tests but that will need further de-bugging

manifests/tasks/retrieve-build-image-info/retrieve-build-image-info.yaml
```diff
@@ -57,28 +57,23 @@ spec:
...
+      echo ;
+      echo -n $DEST_IMAGE_URLS | tee $(results.dest-image-url.path) ;
 
-      echo "Contents of ${URLTXT}:"
-      cat ${URLTXT}
   results:
+  - name: source-image-url
+    description: The source image URL that was inspected 
+  - name: dest-image-url
+    description: The destination image URL(s) where the image will be pushed
```
MarianMacik, line 76: This is no longer relevant as we use the url.txt file.

manifests/tasks/retrieve-build-image-info/retrieve-build-image-info.yaml
```diff
@@ -57,28 +57,23 @@ spec:
...
+      # Set source and destination image URLs in results
+      export SOURCE_IMAGE_URL="docker://${DOCKER_IMAGE_REF}"
+      export DEST_IMAGE_URLS=""
+      for target in "$2"; do
+        DEST_IMAGE_URLS="${DEST_IMAGE_URLS} docker://${target}"
       done
+      echo ;
+      echo -n $SOURCE_IMAGE_URL | tee $(results.source-image-url.path) ;
+      echo ;
+      echo -n $DEST_IMAGE_URLS | tee $(results.dest-image-url.path) ;
```
MarianMacik, line 70: This script should be left as before.

manifests/tasks/retrieve-build-image-info/README.md
```diff
@@ -22,3 +22,5 @@ This task returns more detailed info about a model that has just been built and
 * **buildah-version**: The version of buildah used to build the image
 * **image-digest-reference**: The fully qualified image digest reference of the image
 * **target-image-tag-references**: The fully qualified image reference that the image was pushed to (e.g. registry.example.com/my-org/ai-model:1.0-1)
+* **source-image-url**: The source image URL that was inspected
+* **dest-image-url**: The destination image URL where the image will be pushed
```
MarianMacik, line 26: This can be removed.

manifests/pipelines/s3-fetch-pipeline.yaml
```diff
@@ -314,21 +346,31 @@ spec:
...
     workspaces:
-    - name: images-url
+    - name: images_url
       workspace: build-workspace-pv
   - name: skopeo-copy
     params:
+    - name: SOURCE_IMAGE_URL
+      value: $(tasks.retrieve-build-image-info.results.source-image-url)
+    - name: DESTINATION_IMAGE_URL
+      value: $(tasks.retrieve-build-image-info.results.dest-image-url)
```
MarianMacik, line 356: After that, this can be left empty:
```suggestion
    - name: SOURCE_IMAGE_URL
      value: ""
    - name: DESTINATION_IMAGE_URL
      value: ""
```

jackdelahunt:
`git-fetch` e2e test passes but the `s3-fetch` pipeline is failing on the `test-container-deploy` step. I will try to debug this and see if is your changes

adelton:
As a baseline for trying this change, I did run
```
$ oc apply -k manifests/
$ oc create -f examples/tekton/aiedge-e2e/example-pipelineruns/git-fetch.tensorflow-housing.pipelinerun.yaml
```
on the `main` branch checkout and got `fetch-model-git` failing with
```
{"level":"error","ts":1727270196.2212334,"caller":"git/git.go:53","msg":"Error running git [init /workspace/output//model_dir-tensorflow-housing/]: exit status 128\nfatal: cannot mkdir /workspace/output//model_dir-tensorflow-housing/: Permission denied\n","stacktrace":"github.com/tektoncd-catalog/git-clone/git-init/git.run\n\t/go/src/github.com/tektoncd-catalog/git-clone/image/git-init/git/git.go:53\ngithub.com/tektoncd-catalog/git-clone/git-init/git.Fetch\n\t/go/src/github.com/tektoncd-catalog/git-clone/image/git-init/git/git.go:88\nmain.main\n\t/go/src/github.com/tektoncd-catalog/git-clone/image/git-init/main.go:52\nruntime.main\n\t/usr/lib/golang/src/runtime/proc.go:271"}
{"level":"fatal","ts":1727270196.221359,"caller":"git-init/main.go:53","msg":"Error fetching git repository: exit status 128","stacktrace":"main.main\n\t/go/src/github.com/tektoncd-catalog/git-clone/image/git-init/main.go:53\nruntime.main\n\t/usr/lib/golang/src/runtime/proc.go:271"}
```

With this PR, `fetch-model-git` passes, and so does the rest of the pipeline apart from the push to registry, which was due to me not having credentials set right.

So on the functionality front, this is a big improvement as it actually restores a working state.

manifests/pipelines/s3-fetch-pipeline.yaml
```diff
@@ -318,15 +350,21 @@ spec:
...
+      resolver: git
+      params:
+        - name: url
+          value: https://github.com/openshift-pipelines/tektoncd-catalog
+        - name: revision
+          value: p
+        - name: pathInRepo
+          value: tasks/task-skopeo-copy/0.4.2/task-skopeo-copy.yaml
     workspaces:
     - name: images-url
```
MarianMacik, line 369:
```suggestion
    - name: images_url
```
As per [documentation](https://github.com/openshift-pipelines/tektoncd-catalog/tree/p/tasks/task-skopeo-copy/0.4.2#workspace).
This now needs to be reverted back as we use the new task and not the ClusterTask.

manifests/pipelines/git-fetch-pipeline.yaml
```diff
@@ -343,15 +381,21 @@ spec:
...
+        - name: pathInRepo
+          value: tasks/task-skopeo-copy/0.4.2/task-skopeo-copy.yaml
     workspaces:
     - name: images-url
```
MarianMacik, line 400:
```suggestion
    - name: images_url
```

manifests/pipelines/git-fetch-pipeline.yaml
```diff
@@ -343,15 +381,21 @@ spec:
...
   - name: skopeo-copy
     params:
-    - name: srcTLSverify
-      value: "true"
-    - name: destTLSverify
-      value: "true"
+    - name: SOURCE_IMAGE_URL
+      value: ""
+    - name: DESTINATION_IMAGE_URL
+      value: ""
```
MarianMacik, line 387: Same here.
```suggestion
    - name: SRC_TLS_VERIFY
      value: "true"
    - name: DEST_TLS_VERIFY
      value: "true"
```

manifests/pipelines/s3-fetch-pipeline.yaml
```diff
@@ -318,15 +350,21 @@ spec:
...
   - name: skopeo-copy
     params:
-    - name: srcTLSverify
-      value: "true"
-    - name: destTLSverify
-      value: "true"
+    - name: SOURCE_IMAGE_URL
+      value: ""
+    - name: DESTINATION_IMAGE_URL
+      value: ""
```
MarianMacik, line 356: Now we can leave this as it was before - just explicitly mentioning TLSverify as true. Just name of the [parameters](https://github.com/openshift-pipelines/tektoncd-catalog/tree/p/tasks/task-skopeo-copy/0.4.2#params) are `SRC_TLS_VERIFY` and `DEST_TLS_VERIFY`.
```suggestion
    - name: SRC_TLS_VERIFY
      value: "true"
    - name: DEST_TLS_VERIFY
      value: "true"
```

manifests/tasks/retrieve-build-image-info/retrieve-build-image-info.yaml
```diff
@@ -4,6 +4,9 @@ metadata:
   name: retrieve-build-image-info
 spec:
   description: This task returns more detailed info about a model that has just been built and builds a url.txt file with all image tags to be pushed to
+  workspaces:
+  - name: images-url
```
MarianMacik, line 8:
```suggestion
  - name: images_url
```

MarianMacik (changes requested):
It is failing because the workspace in the new `skopeo-copy` task is called `images_url`.

manifests/pipelines/s3-fetch-pipeline.yaml
biswassri: Ah okay, makes sense! I was looking into logs but couldn't find anything. I'll update these changes.

manifests/pipelines/s3-fetch-pipeline.yaml
biswassri: @MarianMacik Sure. I removed it because these are by default true. But i can add them.

manifests/pipelines/s3-fetch-pipeline.yaml
MarianMacik: @biswassri You can see the log [here](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/opendatahub-io_ai-edge/290/pull-ci-opendatahub-io-ai-edge-main-test-ai-edge/1841494657213665280/artifacts/test-ai-edge/test-ai-edge/artifacts/s3-fetch-bike-rentals-auto-ml-s78w6/). 
Search for `failed` in the taskRuns.txt.

biswassri:
@MarianMacik Resolved all the above requested changes. Should be good to merge after openshift-ci issues resolve.

adelton:
> 
> But I don't think it is caused by this PR as it does not touch it. Will investigate some more.

I've reproduced the issue on the main branch as well.

I had to do
```
$ oc apply -f - < examples/tekton/aiedge-e2e/test-data/tensorflow-housing-test-data-cm.yaml
```
to workaround the issue. It is weird because we don't have this documented in https://github.com/opendatahub-io/ai-edge/blob/main/manifests/README.md so in the past it was not needed as separate step.

adelton:
The next issue I'm hitting is
```
pod "tensorflow-housing-1-6d6c48cd77-cwp82" deleted
service "tensorflow-housing-1" deleted
deployment.apps "tensorflow-housing-1" deleted
Warning: apps.openshift.io/v1 DeploymentConfig is deprecated in v4.14+, unavailable in v4.10000+
Error from server (Forbidden): applications.app.k8s.io is forbidden: User "system:serviceaccount:q-1:pipeline" cannot list resource "applications" in API group "app.k8s.io" in the namespace "q-1"
```
for the git-fetch-tensorflow-housing-wgdvk-stop-deployment Task. But as above, I also see this on the main branch for the stop-deployment ClusterTask, so not specific to this PR.

manifests/pipelines/git-fetch-pipeline.yaml
```diff
@@ -339,21 +377,31 @@ spec:
...
-    - name: images-url
+    - name: images_url
       workspace: build-workspace-pv
   - name: skopeo-copy
     params:
-    - name: srcTLSverify
+    - name: SOURCE_IMAGE_URL
+      value: ""
+    - name: DESTINATION_IMAGE_URL
+      value: ""
```
MarianMacik, line 387: I would remove these as they don't add any value and the values are default anyway.

MarianMacik (commented):
Last thing @biswassri.
It is not needed to properly function but I think we can change it to as it was before to reduce the number of changes, wdyt?

MarianMacik:
> The next issue I'm hitting is now
> 
> ```
> pod "tensorflow-housing-1-6d6c48cd77-cwp82" deleted
> service "tensorflow-housing-1" deleted
> deployment.apps "tensorflow-housing-1" deleted
> Warning: apps.openshift.io/v1 DeploymentConfig is deprecated in v4.14+, unavailable in v4.10000+
> Error from server (Forbidden): applications.app.k8s.io is forbidden: User "system:serviceaccount:q-1:pipeline" cannot list resource "applications" in API group "app.k8s.io" in the namespace "q-1"
> ```
> 
> for the git-fetch-tensorflow-housing-wgdvk-stop-deployment Task when the `oc delete all --selector=app=tensorflow-housing-1` command is run. But as above, I also see this on the main branch for the stop-deployment ClusterTask, so not specific to this PR.

@adelton Yes, you are right. We have found this issue as well when using delete option of the `upon-end` parameter. We use [`keep`](https://github.com/opendatahub-io/ai-edge/blob/main/test/shell-pipeline-tests/openvino-tensorflow-housing/pipelines-test-openvino-tensorflow-housing.sh#L43) in PR check to obtain the logs of the deployment.

MarianMacik:
> > But I don't think it is caused by this PR as it does not touch it. Will investigate some more.
> 
> I've reproduced the issue on the main branch as well.
> 
> I had to do
> 
> ```
> $ oc apply -f - < examples/tekton/aiedge-e2e/test-data/tensorflow-housing-test-data-cm.yaml
> ```
> 
> to workaround the issue. It is weird because we don't have this documented in https://github.com/opendatahub-io/ai-edge/blob/main/manifests/README.md so in the past it was not needed as separate step.

This is documented [here](https://github.com/opendatahub-io/ai-edge/tree/main/manifests#data-for-testing-the-model-inferencing-endpoint).

biswassri:
@MarianMacik The PR check finally passed. Let me know if think any other changes are required.

MarianMacik (approved):
/approve

PR #291, opened 2024-09-03
Increase timeout for the s3-fetch pipeline with bike-rentals

MarianMacik:
## Description
Periodic job finishes with 190-194 tries out of 200 which means there is not enough room in case the environment is slower, leading to false negatives. Therefore, I am raising it to 220. Other pipelines are OK.

## How Has This Been Tested?
Not needed.

## Merge criteria:


biswassri (approved):
/lgtm

